PR #5, opened 2016-04-03
Basic nuget package dependency support.

zaphar:
I've tested in a small workspace and was able to build a binary with a nuget package dependency.

It's probably still rough around the edges a bit but it's a start.

I couldn't find the gerrit repo to push to for code review so feel free to tell me where I should be putting this instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ load(
     "csharp_configure",
     nuget_package
 )
+
 csharp_repositories()
 csharp_configure()
 
@@ -82,6 +83,20 @@ csharp_nunit_test(
 )
 ```
 
+### nuget\_repository
+
+In the WORKSPACE file for your project record a nuget dependency like so.
+This is a repository rule so it will not work unless it is in a workspace
+file.
+
+```python
+nuget_package(
+    name="ndesk_options",
+    package="NDesk.Options",
+    version="0.2.1",
+)
+```
+
 ## Things still missing:
 
 - Handle .resx files correctly.
@@ -91,8 +106,8 @@ csharp_nunit_test(
 
 ## Future nice to haves:
 
+- nuget_packages repository rule that will handle multiple different nuget packages in one rule.
 - Building csproj and sln files for VS and MonoDevelop.
-- Nuget Packaging
 - Windows .NET framwork support
 
 <a name="csharp_library"></a>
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -32,14 +32,22 @@ load(
...
     "csharp_repositories",
     "csharp_configure",
+    "nuget_package",
 )
 csharp_repositories()
 csharp_configure()
+
+nuget_package(
+  name="some_name",
+  pkg_name="Some.Package",
```
davidzchen: Would it be better if we name the attributes `package` and `version` rather than `pkg_name` and `pkg_version`?

README.md
```diff
@@ -32,14 +32,22 @@ load(
...
+  pkg_version="0.1.2",
```
davidzchen: Nit (buildifier style): add spaces before and after `=`

dotnet/csharp.bzl
```diff
@@ -407,6 +417,104 @@ def csharp_configure():
...
+  inputs = set(ctx.files.srcs)
+  return struct(
+    name = ctx.label.name,
+    target_type = ctx.attr._target_type,
+    out = inputs,
+    dlls = inputs,
+    transitive_dlls = set([]),
+  )
+
+nuget_dep = rule(
```
davidzchen: This seems similar to the [java_import](http://bazel.io/docs/be/java.html#java_import) rule. Do you think this rule might be useful for use cases other than nuget packages?

dotnet/csharp.bzl
```diff
@@ -407,6 +417,104 @@ def csharp_configure():
...
+tell it where your nuget.exe is located. You may want to manage that binary
+in your repository as a result.
+
+Args:
+  nuget_package_source: list of sources to use for nuget package feeds.
+  pkg_name: name of the nuget package.
+  pkg_version: version of the nuget package (e.g. 0.1.2)
+"""
+
+NUNIT_BUILD_FILE = """
```
davidzchen: Is this change intentional? I moved the BUILD file for nunit into a standalone BUILD file in f30f2f80984cea9cb53aa46ceacbc7c4046c2f89.

davidzchen:
Awesome! Thanks for working on this!

It looks good for the most part. I have added some comments to your patch.

Also, can you rebase and squash your commits?

damienmg:
(note: a new feature on github: you can squash commits when merging a PR,
you can even force to be the default \o/)

README.md
zaphar: I'm not sure what the prevailing style is here. I have no strong preferences myself. Does the bazel community have a style for this sort of name?

dotnet/csharp.bzl
zaphar: Without it my csharp_test outside of this repository wouldn't build claiming it couldn't find the nunit.BUILD file. If there is a better way to fix that error I'm all for it.

zaphar:
Sure no problem, I'll squash the commits once I've satisfied all the comments I like to save that operation for last in case I need to rollback single changes or something.

dotnet/csharp.bzl
zaphar: It certainly might. I haven't looked at the java_import rule. It's basically just a way to reference other dlls. perhaps it should just be called dll_import or dotnet_import?

It's also possible that some refactoring of the csharp_* rules could make this thin wrapper unnecessary and we could just use filegroups. This was the minimal I could do to make it work.

zaphar:
Okay renamed pkg_name, pkg_version and nuget_package_source and fixed the style nits.

Let me know if you want me to remove the NUNIT_BUILD_FILE piece before I squash things down. It's not really related specifically to this change it just made it a bit easier to test some things.

davidzchen:
Great, thanks!

Just wondering, did you run into any issues with having the nunit build file in a separate BUILD file while you were testing your change? If so, then I would be fine with keeping the NUNIT_BUILD_FILE change.

AFAIU, the `build_file_content` attribute was originally added as a workaround for some of the issues that we had with getting the rules to refer to the correct workspace (see bazelbuild/bazel#1075) but now that the issue is fixed and the rules have been migrated into separate repositories, we can switch back to using `build_file` instead.

zaphar:
I decided to remove the build file change to keep the change targeted better. The fix can be implemented in a separate change. I think it's because the build file is referenced relatively not with a fully qualified name in the external repository.

If I got my git commands correct then this should now be squashed with the nunit build file stuff removed.

davidzchen:
LGTM

Awesome. Thanks!

For the build_file fix, I think the proper fix is to use a label for `build_file` but let's do that in a separate change as you suggested.

kchodorow:
This is broken, it re-defines csharp_configure a second time (see https://github.com/bazelbuild/rules_dotnet/blob/master/dotnet/csharp.bzl#L412 and https://github.com/bazelbuild/rules_dotnet/blob/master/dotnet/csharp.bzl#L503).  Please fix or roll back.

davidzchen:
Sorry, missed this. Opened #8 to fix.

PR #7, opened 2016-04-04
Fix csharp library compilation

kchodorow:
This fixes a couple of issues with the rules:

* It includes mcs in the sandbox, so it can be used to compile.
* It fixes the error message if mcs cannot be found, so that it doesn't
  contain "%s"s.
* It simplifies the local_config_csharp repo, since mcs was being mapped
  to a filegroup, then bound, then mapped into an attribute, which seemed
  excessive.

This doesn't actually fully fix ci.bazel.io, since I'm still getting an error building tests about resolving nunit (`error CS0006: Metadata file 'Nunit.Framework' could not be found`) but this is a C# configuration error, so it gets us closer.

### DIFF
--- a/dotnet/csharp.bzl
+++ b/dotnet/csharp.bzl
@@ -153,8 +153,8 @@ def _csc_get_output(ctx):
 
 def _csc_collect_inputs(ctx, extra_files=[]):
   depinfo = _make_csc_deps(ctx.attr.deps, extra_files=extra_files)
-  inputs = set(ctx.files.srcs) + depinfo.dlls + depinfo.transitive_dlls \
-    + [ctx.file.csc]
+  inputs = (set(ctx.files.srcs) + depinfo.dlls + depinfo.transitive_dlls
+      + [ctx.file.csc])
   srcs = [src.path for src in ctx.files.srcs]
   return struct(depinfo=depinfo,
                 inputs=inputs,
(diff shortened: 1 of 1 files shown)


dotnet/csharp.bzl
```diff
@@ -153,7 +153,8 @@ def _csc_get_output(ctx):
 
 def _csc_collect_inputs(ctx, extra_files=[]):
   depinfo = _make_csc_deps(ctx.attr.deps, extra_files=extra_files)
-  inputs = set(ctx.files.srcs) + depinfo.dlls + depinfo.transitive_dlls
+  inputs = set(ctx.files.srcs) + depinfo.dlls + depinfo.transitive_dlls \
```
davidzchen: Nit: Enclose with parens instead of using a line splice? Also the indentation on L157 is off.

davidzchen:
Thanks for fixing this! Just a minor style nit, otherwise LGTM

davidzchen:
+cc @zaphar

dotnet/csharp.bzl
kchodorow: Done.

kchodorow:
Thanks, I'll wait for the other PR to go in, since I don't want it to accidentally revert any of my changes.

kchodorow:
Merged in 45d8ce52881946392d29baa857c9b0ccc092f8bc.

PR #15, opened 2016-04-29
Add a custom rule for downloading the mono package.

zaphar:
Only works for OS X right now. Linux packages will have
to wait till later.


zaphar:
First pass at this just for OS X at the moment.

zaphar:
This is *not* ready to import since it will break linux platforms and only works on OS X. That said I'd like some early feedback before I tackle the linux version of this. Particularly on how to properly use the configuration to figure out which mono instance to use on linux or other platforms.

Should I name the repository something unique to OS X and use config values to choose the right platform specific repository for mono? Or should/can I make the repository rule figure out which mono package to pull down?

damienmg:
I think the change is pretty good. Keeping the same name and have the repository rule figure out the correct one to download is better because a select would download both versions.

Two minor concerns:
  * the download method might increase the load to download.mono.project.
  * don't they have other archive kind that xar so you don't have to shell out? We can also mirror it on our GCS bucket if you prefer to package it in an easier format.

zaphar:
I asked in the official mono irc channel and they said this was basically the only other option. If google is willing to mirror it then that would probably significantly improve this rule. However it adds a maintenance burden on keeping the mirror up to date.

damienmg:
Well we can mirror it, we have to review the change anyway that's almost the same load. Do you want we mirror it? Do you want it to be repackaged in tar or zip?

zaphar:
I'd much prefer to mirror it I think. Tar or zip is fine. I'll put together the archive with the build file and then the rule will get much simpler. Where should I send the file once it's ready?

damienmg:
Humm let me see how we can set that up. Maybe just send it using Google drive?

zaphar:
Okay I've put together a tarball and placed it here: https://drive.google.com/open?id=0Bzc_lt_M3d9FdW9pdDdzc3FpUE0

When you've got it in the appropriate location let me know and I'll update this change with the new rules.

damienmg:
Sorry for taking so long, here the file is mirrored:
http://bazel-mirror.storage.googleapis.com/download.mono-project.com/archive/4.2.3/macos-10-x86/MonoFramework-MDK-4.2.3.4.macos10.xamarin.x86.tar.gz

damienmg:
(I kept the name of the original file, just changed the extension)

damienmg:
Ok sorry for the previous message, it is your file, renamed into the original file name with a tar gz extension

zaphar:
No problem. This isn't my day job so I'm probably taking longer than you on turnaround.

I'm attempting to make this work and bazel is complaining that it can't download the file with the following error output using --logging=6

```____Downloading from http://bazel-mirror.storage.googleapis.com/download.mono-project.com/archive/4.2.3/macos-10-x86/MonoFramework-MDK-4.2.3.4.macos10.xamarin.x86.tar.gz: 0B
ERROR: java.io.IOException: Error downloading from http://bazel-mirror.storage.googleapis.com/download.mono-project.com/archive/4.2.3/macos-10-x86/MonoFramework-MDK-4.2.3.4.macos10.xamarin.x86.tar.gz to /private/var/tmp/_bazel_jwall/a4f05aced329c5d1ca55a18062777a02/external/mono/mono.tar.gz: Error downloading http://bazel-mirror.storage.googleapis.com/download.mono-project.com/archive/4.2.3/macos-10-x86/MonoFramework-MDK-4.2.3.4.macos10.xamarin.x86.tar.gz to /private/var/tmp/_bazel_jwall/a4f05aced329c5d1ca55a18062777a02/external/mono/mono.tar.gz/MonoFramework-MDK-4.2.3.4.macos10.xamarin.x86.tar.gz: /private/var/tmp/_bazel_jwall/a4f05aced329c5d1ca55a18062777a02/external/mono/mono.tar.gz/MonoFramework-MDK-4.2.3.4.macos10.xamarin.x86.tar.gz (No such file or directory).```

**Comment by zaphar** [comment]
Looks like my issue is caused by https://github.com/bazelbuild/bazel/issues/1235. Using the workaround there.

**Comment by zaphar** [comment]
@damienmg 

So apparently I accidentally uploaded an archive with a broken mcs script. Didn't realize it till I tried to test the download. Can you upload this archive instead? https://drive.google.com/open?id=0Bzc_lt_M3d9FZ19NTmowLTJ0bms

**Comment by damienmg** [comment]
Ok updated.

**Comment by zaphar** [comment]
I'm getting 403 access denied when trying to download.

xml response looks like this.
```xml
<Error>
<Code>AccessDenied</Code>
<Message>Access denied.</Message>
<Details>
Anonymous callers do not have storage.objects.get access to object bazel-mirror/download.mono-project.com/archive/4.2.3/macos-10-x86/MonoFramework-MDK-4.2.3.4.macos10.xamarin.x86.tar.gz.
</Details>
</Error>
```

damienmg:
Ok sorry for that. fixed

zaphar:
Okay updated the rules with the new binary package.

zaphar:
Okay I think this is ready for another look.

It *should* intelligently handle things on OSX and fallback to using a locally installed mono on other platforms. It also allows you to override the download and force the use of a locally installed mono if desired.

README.md
```diff
@@ -49,8 +47,10 @@ nuget_package(
 ```
 
 The `csharp_repositories` rule fetches external dependencies, namely the NUnit
-binaries, and the `csharp_configure` rule sets up the rules to use the local
-Mono C# toolchain installed on your system.
+binaries and the mono repository. Setting `use_local_mono` to True will use
+your installed mono framework instead of downloading one. If you are on OS X
+you can set use_local_mono to False and mono will be downloaded for you by
+bazel. Support for downloading mono on Linux and Windows is coming soon.
```
damienmg: Bazel does not work on Windows yet :)

dotnet/csharp.bzl
```diff
@@ -474,20 +475,51 @@ csharp_autoconf = repository_rule(
...
-  """
-  csharp_autoconf(name = "local_config_csharp")
+def _mono_osx_repository_impl(repository_ctx):
+  # TODO(jwall): The below is necessary due to bug:
+  #   https://github.com/bazelbuild/bazel/issues/1235.
+  #   when that bug is fixed the below should be removed.
+  working_dir = repository_ctx.path("")
+  repository_ctx.file(repository_ctx.path("empty"))
+  if not working_dir.exists:
+    fail("working_dir %s does not exist! See https://github.com/bazelbuild/bazel/issues/1235" % working_dir)
```
damienmg: These two lines should not be necessary

dotnet/csharp.bzl
```diff
@@ -474,20 +475,51 @@ csharp_autoconf = repository_rule(
...
+
+  download_output = repository_ctx.path("")
+  #download the package
```
damienmg: nit: space after #

dotnet/csharp.bzl
```diff
@@ -474,20 +475,51 @@ csharp_autoconf = repository_rule(
...
+  repository_ctx.download_and_extract(
+    repository_ctx.attr._osx_pkg,
+    download_output,
+    repository_ctx.attr._osx_sha256,
+    "", "mono")
+
+  # now we create the build file.
+  toolchain_build = """
+package(default_visibility = ["//visibility:public"])
+exports_files(["mono", "mcs"])
```
damienmg: nit: on Linux there is sandboxing and you might need to declare a dependency on all the files of the framework. Not needed yet so you might ignore it.

dotnet/csharp.bzl
```diff
@@ -474,20 +475,51 @@ csharp_autoconf = repository_rule(
...
+"""
+  repository_ctx.file("bin/BUILD", toolchain_build)
+
+def _mono_repository_impl(repository_ctx):
+  print("os name: %s" % repository_ctx.os.name)
```
damienmg: This will show a warning, remove.

dotnet/csharp.bzl
```diff
@@ -474,20 +475,51 @@ csharp_autoconf = repository_rule(
...
+  if repository_ctx.attr.use_local:
+    _csharp_autoconf(repository_ctx)
+  elif repository_ctx.os.name == "mac os x":
```
damienmg: find("mac") != -1 instead?

dotnet/csharp.bzl
```diff
@@ -474,20 +475,51 @@ csharp_autoconf = repository_rule(
...
+    _mono_osx_repository_impl(repository_ctx)
+  else:
+    #if nothing else works look for a local version of mono
+    _csharp_autoconf(repository_ctx)
+
+mono_package = repository_rule(
+  implementation = _mono_repository_impl,
+  attrs = {
+    "_osx_pkg": attr.string(default="http://bazel-mirror.storage.googleapis.com/download.mono-project.com/archive/4.2.3/macos-10-x86/MonoFramework-MDK-4.2.3.4.macos10.xamarin.x86.tar.gz"),
```
damienmg: why is it a hidden attribute, why not make it a constant if you don't want the user to override it?

dotnet/mono_unpack.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
damienmg: Not used anymore, delete?

damienmg:
That's really nice, I still have some comments.

README.md
```diff
@@ -49,8 +47,10 @@ nuget_package(
 ```
 
 The `csharp_repositories` rule fetches external dependencies, namely the NUnit
-binaries, and the `csharp_configure` rule sets up the rules to use the local
-Mono C# toolchain installed on your system.
+binaries and the mono repository. Setting `use_local_mono` to True will use
+your installed mono framework instead of downloading one. If you are on OS X
+you can set use_local_mono to False and mono will be downloaded for you by
```
davidzchen: Nit: back quotes around `True`, `False`, and `use_local_mono` elsewhere in the text.

README.md
```diff
@@ -34,12 +34,10 @@ git_repository(
 load(
     "@io_bazel_rules_dotnet//dotnet:csharp.bzl",
     "csharp_repositories",
-    "csharp_configure",
     nuget_package
 )
 
-csharp_repositories()
-csharp_configure()
+csharp_repositories(use_local_mono=True)
```
davidzchen: Nit: BUILD convention uses spaces around the `=` for keyword args.

davidzchen:
This is awesome! Thanks for putting this together.

dotnet/csharp.bzl
zaphar: I'll cross that boat when I come to it for Linux I think. Declaring that dependency could be a bit of a challenge.

dotnet/csharp.bzl
zaphar: I wasn't clear on the expected style here. In some places I've seen it done this way and in others I've seen it as a constant. I don't have a preference myself. I'll change them to a constant instead.

dotnet/mono_unpack.sh
zaphar: Whoops I meant to do that and forgot.

zaphar:
Okay I think I addressed all the comments.

damienmg:
LGTM

PR #20, opened 2016-09-07
Fix build on Linux

kchodorow:
This doesn't fix Jenkins, but it does let you do `bazel build //...` on Linux. I was initially getting the error:

```
error CS0006: Metadata file `Nunit.Framework' could not be found`
```

The file's name is nunit.framework.xml, so AFAICT on a case-sensitive filesystems, it was looking for Nunit.Framework.xml.  Then it couldn't find nunit.framework.xml anyway, since the nunit_framework target only included the .dlls (not the .xml file).


davidzchen:
LGTM

zaphar:
Sorry it took so long. I've been busy with a job search and my time was pretty limited.

Issue #14, opened 2016-04-28
Pull in mono as an external workspace.


zaphar:
This is going to require platform specific repositories probably.

On OS-X we can probably reuse some of this code: https://github.com/mono/release/blob/master/defs/managed-components#L52

For Linux we can probably unpack one of the distro packages.

For MS I'm not sure what we should do but that can probably wait till bazel windows support is stabilized.

tomaszstrejczek:
I couldn't find any downloadable working in-place (without installing) mono installation for Windows or Linux. 
A similar problem exists for .NET.

purkhusid:
Cleaning up. Going to close this for now since Mono is out of scope at the moment.

PR #21, opened 2016-09-07
Fix cd error in test

kchodorow:
(See the Fix cd commit description.)

### DIFF
--- a/dotnet/csharp.bzl
+++ b/dotnet/csharp.bzl
@@ -240,7 +240,7 @@ def _cs_nunit_run_impl(ctx):
       depinfo)
 
   _csc_compile_action(ctx, output, outputs, collected_inputs,
-                      extra_refs=["Nunit.Framework"])
+                      extra_refs=["nunit.framework"])
 
   _make_nunit_launcher(ctx, depinfo, output)
 
--- a/dotnet/csharp.bzl
+++ b/dotnet/csharp.bzl
@@ -100,7 +100,9 @@ def _make_csc_arglist(ctx, output, depinfo, extra_refs=[]):
 _NUNIT_LAUNCHER_SCRIPT = """\
 #!/bin/bash
 
-cd $0.runfiles
+if [[ -e "$0.runfiles" ]]; then
+  cd $0.runfiles/{workspace}
+fi
 
 # TODO(jeremy): This is a gross and fragile hack.
 # We should be able to do better than this.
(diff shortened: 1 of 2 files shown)


kchodorow:
Note that this doesn't actually fix the error that we're getting on Jenkins.  It just fixes [this error](http://ci.bazel.io/job/PR-rules_dotnet/BAZEL_VERSION=latest,PLATFORM_NAME=darwin-x86_64/3/console):

```
/private/var/tmp/_bazel_ci/e305a8db68ed3b3e5540ded576293b6a/execroot/darwin-x86_64/bazel-out/local-fastbuild/bin/examples/example_test/MyTest.runfiles/io_bazel_rules_dotnet/examples/example_test/MyTest: line 3: cd: /private/var/tmp/_bazel_ci/e305a8db68ed3b3e5540ded576293b6a/execroot/darwin-x86_64/bazel-out/local-fastbuild/bin/examples/example_test/MyTest.runfiles/io_bazel_rules_dotnet/examples/example_test/MyTest.runfiles: No such file or directory
```

I'm actually pretty confused why this is failing:

```
____From Testing //examples/example_test:MyTest:
==================== Test output for //examples/example_test:MyTest:
NUnit-Console version 2.6.4.14350
Copyright (C) 2002-2012 Charlie Poole.
Copyright (C) 2002-2004 James W. Newkirk, Michael C. Two, Alexei A. Vorontsov.
Copyright (C) 2000-2002 Philip Craig.
All Rights Reserved.

Runtime Environment - 
   OS Version: Unix 15.5.0.0
  CLR Version: 4.0.30319.17020 ( Mono 4.0 ( 4.2.3 (explicit/832de4b Thu Mar  3 19:24:57 EST 2016) ) )

File type not known: ../nunit/NUnit-2.6.4/bin/framework/nunit.framework.xml
WARNING: The runtime version supported by this application is unavailable.
Using default runtime: v4.0.30319
================================================================================
```

It doesn't look like there are any errors there?

kchodorow:
Jenkins, test this please.

kchodorow:
Okay, so this does work on Darwin without the NUnit.Framework change.  Although the error is different than the "main" CI was getting (`execroot/external/mono/bin/mono: No such file or directory`), so... ?

damienmg:
The problem with Darwin is that it use an external repository to declare it
deps so it just means the compiler isn't shipped in the sandbox. A rough
fix is to add a glob in the build file and ship everything in as an action
dep. I almost got that working last week then got side tracked with the fix
on CI and the Tutorial.

kchodorow:
Is the PR jenkins disabling sandboxing or something?  Because this passes with that...

damienmg:
PR vetting does not use bazel HEAD, only bazel release.

Also we have deactivated linux for dotnet because of #14

damienmg:
So I had some time to work on this breakage today and we cannot fix it.

To fix it, we need to pass a filegroup of the files from the mono toolchain which contains spaces. Therefore those filename cannot be labels therefore cannot be included.

I see if we can just mark the action local for now.

damienmg:
Ok of course this cannot be made because these are skylark action... I see only one solution for now: deactivate sandboxing for this ruleset. That's terribly sad :(

kchodorow:
I think this change is still legit, mind if I merge it?

damienmg:
LGTM 👍

PR #22, opened 2016-09-09
Deactivate sandboxing

damienmg:
Because of spaces in file names of the mono framework, it cannot be
shipped into the sandbox, so deactivate sandboxing for now for this project.

This should make ci.bazel.io green again.

Also remove the extraneous 'Rules' in the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,10 @@
   </ul>
 </div>
 
+__WARNING:__ Theses rules are not compatible with
+[sandboxing](https://bazel.io/blog/2015/09/11/sandboxing.html).
+
+
 ## Overview
 
 This is a minimal viable set of C# bindings for building csharp code with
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -3,7 +3,6 @@
 ## Rules
 
 <div class="toc">
```
davidzchen: Since we are no longer copying this README.md for Bazel.io (as we currently are for the [Packaging rules](https://www.bazel.io/versions/master/docs/be/pkg.html)), we can turn this into a Markdown list.

FYI the main blocker for generating Skydoc documentation for the `rules_*` repos is adding an overview page for the generated documentation, which I am currently working on. Once that is done, then we can generate Skydoc for each of these repos and serve them, similar to how Rust libraries use [generated Rustdoc](http://rust-num.github.io/num/num/index.html) as their project websites.

davidzchen:
LGTM with just a minor comment

PR #29, opened 2016-10-06
Adds RULES_DOTNET_USE_LOCAL_MONO environment variable.

pcj:
- Setting this variable to a truthy value is equivalent to setting
  the use_local_mono=True.

This is useful for CI when you don't want to have to check in the
value of the flag to version control.


bazel-io:

pcj:
@zaphar Can you take a look?  I keep having to amend commits when I forget to remove the use_local_mono=True flag.

zaphar:
Thanks for the work!

PR #30, opened 2016-10-06
Don't ignore extra_refs.

pcj:
1. The current implementation will not add -reference args for
extra_refs if depinfo.refs is empty.  This causes failure for
csharp_nunit_test due to the `Nunit.Framework` reference being
elided. Adding the or clause fixes this bug.

2. De-duplicates library args list.


bazel-io:

pcj:
@zaphar  This one has been blocking me from using the main bazelbuild/rules_dotnet repo rather than my fork (can't seem to use csharp_nunit_test without it).  Just getting around to pushing this upstream.

pcj:
Thanks!  To clarify, the failure ~~is~~ was for simple tests that have no library dependencies.

PR #31, opened 2016-10-06
Hard fail when operating system is unknown.

pcj:


bazel-io:

pcj:
@zaphar Somehow I keep forgetting that (obviously) linux is not supported.  Then I spend time trying to figure out why the build is failing (i.e., stopped excluding a csharp test on linux).

This at least informs the user better.  Let me know if there is a use-case for letting the autoconf try for best effort (which seems covered by use_local_mono=True).

zaphar:
I need to think about this one a bit and I'm a bit time constrained at the moment so sorry for long latency responses here.

pcj:
Sounds good.  The current behavior will try to find a local mono instance on the non-mac host (perhaps from something like apt-get install mono-complete).  This change would require a linux user to say `use_local_mono=True`.  

In the future this would not be an issue if rules_dotnet could download mono for linux.  I have not looked much into how complicated linux support is for rules_dotnet though.

Issue #34, opened 2016-10-27
csharp_binary gives "Too many levels of symbolic links" error on OS X


iainmerrick:
I'm having trouble getting the csharp_binary rule to work. I'm on OS X 10.12.

The rule seems to build OK, but when I run it I get this error:

```
$ bazel run :inklecate
INFO: Found 1 target...
Target //:inklecate up-to-date:
  bazel-bin/inklecate.exe
  bazel-bin/inklecate
INFO: Elapsed time: 0.066s, Critical Path: 0.00s

INFO: Running command line: bazel-bin/inklecate
Cannot open assembly 'inklecate.exe': Too many levels of symbolic links.
ERROR: Non-zero return code '2' from command: Process exited with status 2.
```

The rule looks like this:

```
csharp_binary(
    name = "inklecate",
    main_class = "Ink.CommandLineTool",
    srcs = glob([
        "inklecate/*.cs",
        "inklecate/**/*.cs",
        "ink-engine-runtime/*.cs",
    ]),
    warn = 3, # Suppress documentation warnings
)
```

I've tried both `use_local_mono = True` and `False` but that doesn't seem to make a difference.

Full code is here: https://github.com/iainmerrick/ink

Further discussion on the Bazel mailing list: https://groups.google.com/forum/#!msg/bazel-discuss/M-qFbm0x6mY/KOy_ljc4AAAJ

zaphar:
Interesting, if you execute the compiled assembly manually does it work?

iainmerrick:
Running `./bazel-bin/inklecate` gives the same error.

Running `mono bazel-bin/inklecate.exe` works.

iainmerrick:
Aha! Taking a closer look at the generated `inklecate` script, this looks bad:

    bazel-bin/inklecate.runfiles/__main__$ ls -a -l
    total 16
    drwxr-xr-x  5 iain  wheel  170 27 Oct 20:11 .
    drwxr-xr-x  5 iain  wheel  170 27 Oct 20:11 ..
    drwxr-xr-x  3 iain  wheel  102 27 Oct 20:11 external
    lrwxr-xr-x  1 iain  wheel  114 27 Oct 20:11 inklecate -> /private/var/tmp/_bazel_iain/3748381dda9ed83df4937428344e9aeb/execroot/ink/bazel-out/local-fastbuild/bin/inklecate
[...]

PR #35, opened 2016-10-28
Avoid creating circular symlinks in launch scripts (#34)

iainmerrick:
The launcher script for csharp_binary and csharp_nunit_test rules
creates symlinks (ln -s -f) at the top level of runfiles for its
.exe and other files which normally live in subdirectories.

However, when the .exe was already at the top level, it would be
overwritten with a circular symlink. To fix this we:
- Remove the -f to avoid trashing the real file
- Manually check for the symlink's existence instead

### DIFF
--- a/dotnet/csharp.bzl
+++ b/dotnet/csharp.bzl
@@ -104,10 +104,15 @@ if [[ -e "$0.runfiles" ]]; then
   cd $0.runfiles/{workspace}
 fi
 
+# Create top-level symlinks for lib files.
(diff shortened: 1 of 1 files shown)


bazel-io:

iainmerrick:
Sorry, looks like //examples/example_test:MyTest fails with this commit. I assumed that the tests had been run automatically by the bot. I'll update the patch.

iainmerrick:
Verified that "bazel test ..." works now

Issue #36, opened 2016-11-02
csharp_binary can't be used as a Bazel tool because working directory is changed


iainmerrick:
The launcher script for a `csharp_binary` currently cd's into its runfiles directory. But this means it can't be used as a tool (i.e. in a `genrule`) because Bazel passes file arguments as relative paths. Console output works, but named `srcs` and `outs` don't.

The current working directory needs to be preserved when running a `csharp_binary` as a tool.

iainmerrick:
Verified (just need to build the genrules under //examples/example_tool)

PR #37, opened 2016-11-02
Make csharp_binary work correctly in genrule (#36)

iainmerrick:
The working directory needs to be preserved so that relative paths
work correctly. To fix this, we pushd/popd to set up runfiles
rather than just cd'ing into it.

example_tool tests this with both single-output and multi-output
genrules.


bazel-io:

PR #45, opened 2017-11-14
Use ctx.actions.run rather than ctx.action

j3parker:
This is important/useful for Windows where we are more inclined to avoid bash.

This is particularly important in light of https://github.com/bazelbuild/bazel/issues/4013 which for me results in extremely slow builds on Windows (I'm running a fork of this repo with .NET framework support; I'll be upstreaming the simple platform-agnostic changes to start.)

`ctx.action` was recently deprecated; the details are [in the docs](https://docs.bazel.build/versions/master/skylark/lib/ctx.html#action).


bazel-io:

PR #46, opened 2017-11-14
(Minor) Always pass add_mono=True to _cs_runfiles.

j3parker:
This matches the current behaviour and gets rid of a confusing unused
local variable. I don't know what the original plan was for making this
optional/default False (maybe cross-platform ideas?) but at least this
is clearer.

The `add_mono` param was added by @zaphar in https://github.com/bazelbuild/rules_dotnet/commit/838e7ea0d535ca4bb3ec4704796db653a67262f4


bazel-io:

PR #47, opened 2017-11-28
Use param files for the compiler

j3parker:
Windows has a max command-line length of around 32k bytes. The default
for Args is to spill to param files if the length of the executable,
params and spaces between them sum to that number. Unfortunately the
math doesn't work out exactly like that on Windows (e.g. there is also
a null terminator.) For now, setting use_always to True is the
conservative option. Long command lines are probable with C# due to
organizing files by namespace.

See https://github.com/bazelbuild/bazel/issues/4083

CC @zaphar

### DIFF
--- a/dotnet/csharp.bzl
+++ b/dotnet/csharp.bzl
@@ -48,19 +48,24 @@ def _get_libdirs(dlls, libdirs=[]):
   return [dep.dirname for dep in dlls] + libdirs
 
 def _make_csc_arglist(ctx, output, depinfo, extra_refs=[]):
+  args = ctx.actions.args()
+
   flag_start = ctx.attr._flag_start
-  args = [
-       # /out:<file>
-      _make_csc_flag(flag_start, "out", output.path),
-       # /target (exe for binary, library for lib, module for module)
-      _make_csc_flag(flag_start, "target", ctx.attr._target_type),
-      # /fullpaths
-      _make_csc_flag(flag_start, "fullpaths"),
-      # /warn
-      _make_csc_flag(flag_start, "warn", str(ctx.attr.warn)),
-      # /nologo
-      _make_csc_flag(flag_start, "nologo"),
-  ]
+
+  # /out:<file>
+  args.add(_make_csc_flag(flag_start, "out", output.path))
+
+  # /target (exe for binary, library for lib, module for module)
+  args.add(_make_csc_flag(flag_start, "target", ctx.attr._target_type))
+
+  # /fullpaths
+  args.add(_make_csc_flag(flag_start, "fullpaths"))
+
+  # /warn
+  args.add(_make_csc_flag(flag_start, "warn", str(ctx.attr.warn)))
+
+  # /nologo
+  args.add(_make_csc_flag(flag_start, "nologo"))
 
   # /modulename:<string> only used for modules
   libdirs = _get_libdirs(depinfo.dlls)
@@ -68,22 +73,23 @@ def _make_csc_arglist(ctx, output, depinfo, extra_refs=[]):
 
   # /lib:dir1,[dir1]
   if libdirs:
-    args += [_make_csc_flag(flag_start, "lib", ",".join(list(libdirs)))]
+    args.add(_make_csc_flag(flag_start, "lib", ",".join(list(libdirs))))
 
   # /reference:filename[,filename2]
   if depinfo.refs or extra_refs:
-    args += [_make_csc_flag(flag_start, "reference",
-                            ",".join(list(depinfo.refs + extra_refs)))]
+    args.add(_make_csc_flag(flag_start, "reference",
+                            ",".join(list(depinfo.refs + extra_refs))))
   else:
-    args += extra_refs
+    args.add(extra_refs)
 
   # /doc
   if hasattr(ctx.outputs, "doc_xml"):
-    args += [_make_csc_flag(flag_start, "doc", ctx.outputs.doc_xml.path)]
+    args.add(_make_csc_flag(flag_start, "doc", ctx.outputs.doc_xml.path))
 
   # /debug
   debug = ctx.var.get("BINMODE", "") == "-dbg"
-  args += [_make_csc_flag(flag_start, "debug")] if debug else []
+  if debug:
+    args.add(_make_csc_flag(flag_start, "debug"))
 
   # /warnaserror
   # TODO(jeremy): /define:name[;name2]
@@ -91,7 +97,7 @@ def _make_csc_arglist(ctx, output, depinfo, extra_refs=[]):
 
   # /main:class
   if hasattr(ctx.attr, "main_class") and ctx.attr.main_class:
-    args += [_make_csc_flag(flag_start, "main", ctx.attr.main_class)]
+    args.add(_make_csc_flag(flag_start, "main", ctx.attr.main_class))
 
   # TODO(jwall): /parallel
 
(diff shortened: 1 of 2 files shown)


bazel-io:

dotnet/csharp.bzl
```diff
@@ -48,50 +48,56 @@ def _get_libdirs(dlls, libdirs=[]):
   return [dep.dirname for dep in dlls] + libdirs
 
 def _make_csc_arglist(ctx, output, depinfo, extra_refs=[]):
+  args = ctx.actions.args()
```
j3parker: https://docs.bazel.build/versions/master/skylark/lib/Args.html

j3parker:
Docs for response files: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/response-file-compiler-option

(Also supported by mcs)

j3parker:
Bump?

dotnet/csharp.bzl
```diff
@@ -48,50 +48,56 @@ def _get_libdirs(dlls, libdirs=[]):
...
+
   flag_start = ctx.attr._flag_start
```
laszlocsomor: What's the point of having the `_flag_start` attribute if it's hidden (starts with "_") so the value is constant and the user needs to remember passing it to `_make_csc_flag` and it's always concatenated with the flag name? Wouldn't it be easier, more readable, and faster to use e.g. "-out" instead of "out" as flag names?

I won't block an LGTM on this issue but I think it'd be useful to fix.

dotnet/csharp.bzl
```diff
@@ -48,50 +48,56 @@ def _get_libdirs(dlls, libdirs=[]):
...
+  args.add(_make_csc_flag(flag_start, "nologo"))
 
   # /modulename:<string> only used for modules
   libdirs = _get_libdirs(depinfo.dlls)
   libdirs = _get_libdirs(depinfo.transitive_dlls, libdirs)
 
   # /lib:dir1,[dir1]
   if libdirs:
-    args += [_make_csc_flag(flag_start, "lib", ",".join(list(libdirs)))]
+    args.add(_make_csc_flag(flag_start, "lib", ",".join(list(libdirs))))
```
laszlocsomor: You can join these lazily and save analysis-time memory:

```
def _comma_join(flag, values):
  return _make_csc_flag(None, flag, ",".join(list(libdirs)))

(...)

  if libdirs:
    args.add((flag_start + "lib", libdirs), map_fn=_comma_join)
```

dotnet/csharp.bzl
```diff
@@ -48,50 +48,56 @@ def _get_libdirs(dlls, libdirs=[]):
...
 
   # /reference:filename[,filename2]
   if depinfo.refs or extra_refs:
-    args += [_make_csc_flag(flag_start, "reference",
-                            ",".join(list(depinfo.refs + extra_refs)))]
+    args.add(_make_csc_flag(flag_start, "reference",
+                            ",".join(list(depinfo.refs + extra_refs))))
```
laszlocsomor: opt: you can join this lazily too

dotnet/csharp.bzl
j3parker: Awesome! Will do x2

dotnet/csharp.bzl
j3parker: Hm, I think the original intent was a nod to supporting Windows: `csc` expects its flags to start with `/` rather than `-`. Ultimately this would become an attr of a toolchain.

Some quick googling led to: https://manpages.debian.org/wheezy/mono-mcs/mcs.1.en.html
> The Mono C# compiler accepts the same command line options that the Microsoft C# compiler does. Those options can start with a slash or a dash (/checked is the same as -checked).

I could confirm that on Linux/mono and then "inline" `/` into `_make_csc_flag` for simplicity. Does that sound good? As a follow-up PR?

laszlocsomor: Sure. Thanks!

dotnet/csharp.bzl
j3parker: I went with separate args instead because I had trouble with doing this (join all with commas and prefix that with `/foo:`) with a scalar `map_fn` (I think that's what you were going for in your example - where the value is a tuple.)

`before_each` won't work because it inserts a separate argument between each argument, but csc/mcs wants this to be a prefix of each argument (e.g. no space/newline)

`format` is simple to read but does result in larger param files. In practice the number of `lib` and `reference` args shouldn't be too crazy. In our build we have ~1500 `csharp_library` targets and with this change our total param file size went from around 13.1MB (lol) to 13.9MB.

dotnet/csharp.bzl
```diff
@@ -48,50 +48,55 @@ def _get_libdirs(dlls, libdirs=[]):
...
+  args.add(_make_csc_flag(flag_start, "nologo"))
 
   # /modulename:<string> only used for modules
   libdirs = _get_libdirs(depinfo.dlls)
   libdirs = _get_libdirs(depinfo.transitive_dlls, libdirs)
 
   # /lib:dir1,[dir1]
   if libdirs:
-    args += [_make_csc_flag(flag_start, "lib", ",".join(list(libdirs)))]
+    args.add(libdirs, format=flag_start + "lib:%s")
```
j3parker: I kept `flag_start` for consistency but didn't use `_make_csc_flag` (which otherwise deals with `:`) in the next PR I'll switch to `/` across the board and drop `_make_csc_flag` because it's not pulling it's weight.

laszlocsomor: Thanks! Did you try if it works?
For what I can tell, this looks good.

j3parker: Yep, sorry forgot to mention: tried it on Mac and Linux with Mono.

On that note: what would be the process for setting up CI for this repo? What is the preferred service for that (Travis? A Jenkins service you run? etc.) and would it work if I PR'd the stuff assuming a maintainer will click the buttons I can't to turn stuff on?

I can look around at other rules repos to get a sense of how to organize the tests but if this is something that makes sense for contribution and you have pointers that's cool too.

j3parker: Oh neat, I didn't notice. Is it set up to check PRs? I don't see any checks in this PR but maybe it only posts on fail?

j3parker: Judging by https://ci.bazel.io/job/rules_dotnet/GitHubPollLog/ it's set to trigger on master chages only. I'd be cool to have it in PRs so these changes we're safer to review.

laszlocsomor: Honestly, I don't know. If I understand the Jenkins UI correctly, there's a [directory for PR jobs](https://ci.bazel.io/view/Bazel%20bootstrap%20and%20maintenance/job/PR/), which I suppose are the tests that Jenkins runs on a PR or Gerrit review. But all runs are weeks or months old, so this might be turned of, or not running as presubmit, or both.

Let me pull in @philwo who started working on CI today and might know more.

laszlocsomor: Actually, maybe I'm wrong and this CI job is supposed to test Bazel changes with the `rules_dotnet` depot, not the changes in the `rules_dotnet` depot itself.

philwo: I guess this was supposed to work, but maybe it doesn't? It surely would be very useful. I wonder if we can just trigger it using the magic words? Let me try this.

I'm working on improving our CI, but it will be a few weeks until we have something to show, as the only person who built and maintained the current system left the team recently and we have to figure out how this all works first.

philwo:
Jenkins, test this please.

j3parker:
Do I have to do anything else to get this merged?

bazel-io:

j3parker:
Oooo fancy, that triggered CI :)

j3parker:
Thanks @laszlocsomor !

PR #49, opened 2018-01-12
Hard-code / as the flag start

j3parker:
This should work cross-platform (mono and otherwise.) This is important
for non-mono support on Windows.

https://manpages.debian.org/wheezy/mono-mcs/mcs.1.en.html

"The Mono C# compiler accepts the same command line options that the
Microsoft C# compiler does. Those options can start with a slash or a
dash (/checked is the same as -checked)."

This was tested on Linux and Mac with Mono.

Follow-up to comments from #47


bazel-io:

PR #50, opened 2018-01-15
Support embedded resources in csharp_* rules

j3parker:
Notes

* I'm assuming #49 will get merged and not using `_make_csc_flag` when I `args.add` here. This PR **doesnt** depend on that one, however.
* It would probably be a good idea to ban `.resx` from the `resources` attribute (I think I'd have to throw an error in the impl?) but I'll explain this a bit more in an larger issue I'm going to open about resx files shortly (EDIT: #51).
* I tested this on Linux+Mono (I'm porting these changes from a Windows+.NET framework fork) and the output of `strings` on the `MyClass` DLL contains the text file (and the param file that is generated looks correct.)

### DIFF
--- a/dotnet/csharp.bzl
+++ b/dotnet/csharp.bzl
@@ -98,6 +98,8 @@ def _make_csc_arglist(ctx, output, depinfo, extra_refs=[]):
   if hasattr(ctx.attr, "main_class") and ctx.attr.main_class:
     args.add(_make_csc_flag(flag_start, "main", ctx.attr.main_class))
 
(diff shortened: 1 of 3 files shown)


bazel-io:

j3parker:
Docs for `/resource`: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/resource-compiler-option

Not setting the identifier (it defaults to the file name) or the accessibility modifier (it defaults to `public`). This matches the behavior of msbuild+csproj with `EmbeddedResource` or `Content` items from what I can tell.

Issue #33, opened 2016-10-12
Release 0.0.3


pcj:
With the cfg parameter update, consider cutting 0.0.3 release as this requires bazel 0.3.1 or above.

mattmoor:
@steren @damienmg @kchodorow The [setup](https://github.com/bazelbuild/rules_dotnet#setup) instructions actually reference a 0.0.3 release tag, which doesn't exist.

PR #52, opened 2018-01-21
(Skylint) Use dicts.add rather than deprecated +

j3parker:
Not urgent, but following the directions from the Skylint docs: https://github.com/bazelbuild/bazel/blob/7f86d4a34bbd76295b8753ffc2740e43aa228fc2/site/docs/skylark/skylint.md#using-the-operator--on-dictionaries-deprecated-plus-dict

A few questions:

0. Is this worth the dependency on skylib at all?
1. I assume it's desirable to have a standard name for importing something like skylib. "bazel_skylib" is mentioned by skylint docs (linked above) and is the name used in skylib's `WORKSPACE` file so I chose it here. This library uses "io_bazel_rules_dotnet" to refer to itself which looks like an attempt at a URI-based namespacing convention. Are there plans for conventions around naming external workspaces?
2. Is using a specific commit in the `http_archive` for skylint acceptable for now? (both our `WORKSPACE` and the `README`.)
3. This change isn't backwards compatible. The instructions in the `README` don't line up because it is pulling v0.3 of rules_dotnet which doesn't have these changes. Should we release a new version? Related: #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,14 @@ might become especially valuable.
 Add the following to your `WORKSPACE` file to add the external repositories:
 
 ```python
+# A newer version should be fine
+http_archive(
+  name = "bazel_skylib",
+  url = "https://github.com/bazelbuild/bazel-skylib/archive/ff23a62c57d2912c3073a69c12f42c3d6e58a957.zip",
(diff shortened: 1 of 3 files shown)


bazel-io:

PR #95, opened 2018-12-26
Add nuget2bazel --skipSha256 option

pcj:
Allows user to opt-out of calculating a sha256 value for generated nuget_package rules.


pcj:
I find the expected sha256 values printed by the nuget2bazel tool rarely match the actual value.  This option allows a user to not print these values.

PR #104, opened 2019-01-30
Create CODEOWNERS

dslomov:
Adding GitHub repo admins as global CODEOWNERS


CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @zaphar @davidzchen
```
laurentlb: David is no longer active.
I think @tomaszstrejczek should be owner instead.

Issue #107, opened 2019-02-08
error to download .Net Core sdk version v2.1.502


nmalsang:
An error happens during the download of the .net core sdk version v2.1.502. It looks like this version is not available. I fix it by using the version v2.1.503. 
I can suggest my fix .

tomaszstrejczek:
The PR would be highly appreciated.

nmalsang:
The PR is coming soon

nmalsang:
I prepare a first fix for the rules for .Net Core and add a unit test on my fork. 
But I have as dependency for my unit test : "@io_bazel_rules_dotnet//dotnet/stdlib.core:mscorlib.dll",
as you can see here : 
  core_binary(
    name = "custom_bin",
    srcs = [
        "main.cs",
    ],
    out = "alt_bin.exe",
    deps = [
        "@io_bazel_rules_dotnet//dotnet/stdlib.core:mscorlib.dll",
    ],

I would like remove this here because mscorlib.dll and its dependencies are standard default libraries that we always need to build a c# file.  The dotnet csc.dll command line needs to reference by default all the dependencies of mscorlib.dll mentioned in the file //dotnet/stdlib.core/macro.bzl. I would like modify the file //dotnet/private/actions/assembly_core.bzl at line 54 where we write the command line to compile to add a loop on the dependencies of mscorlib.dll and write /reference ***.dll in the command line. If I well understood I need to create a depset from the string "@io_bazel_rules_dotnet//dotnet/stdlib.core:mscorlib.dll" but I don't see how to do that. 
Does anyone could help me to do that please? or explain me how to do that?
Thanks in advance

nmalsang:
the code I try to do : 
    




in assembly_core.bzl line 54 : 
std_deps = ???? the dependencies of mscorlib.dll ???
    transitive_std_deps = depset(direct = std_deps, transitive = [d[DotnetLibrary].transitive for d in std_deps])
    args.add_all(transitive_std_deps, format_each = "/reference:%s", map_each = _map_dep)

Any suggestion is welcome.
Thanks

PR #121, opened 2019-04-15
#107 fix context .Net Core and the command line dotnet csc.dll ...

nmalsang:
#107 fix context .Net Core : core_sdk . The command line dotnet csc.dll does not support mscorlib.dll as reference we need to reference the dependencies of mscorlib.dll. Remove the use of the targetframework.cs file for which is not compatible with .Net Core. Add unit test : test/core/core_binary. Modify the list of the native_deps for core_sdk.


tomaszstrejczek:
I tried it on my machine and the build fails:
bazel build //...
...
ERROR: C:/rules_dotnet/tools/nuget2bazel/BUILD:3:2: Compiling tools/nuget2bazel:v2.1.200_nuget2bazel failed (Exit 1)
C:\users\tomek\_bazel_tomek\3cegijzu\execroot\io_bazel_rules_dotnet\tools\nuget2bazel\ProjectBazel.cs(22,32): error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.
...

nmalsang:
I am now able to build nuget2bazel with the version .Net Core v2.1.200 and v2.1.503. 
Let me know if you see other issues. 
For now to make work the unit test //tests/core/core_binary I need to add the dependency:
"@io_bazel_rules_dotnet//dotnet/stdlib.core:mscorlib.dll",
As it is a default base dependency I looking for a way to add it automatically to the command line to avoid to have to write this dependency in the core_binary rules. Does it make sense for you? If yes I have few question to achieve that work. 
Regards

PR #127, opened 2019-05-09
add unit tests for dotnet core rules using xunit as testing framework

nmalsang:
add unit tests for dotnet core rules using xunit as testing framework
add sha256 for xunit repos


nmalsang:
Thanks tomas!! :)

PR #133, opened 2019-06-26
Fix TravisCI and AppVeyor example_binary:hello target names

toqueteos:
I've send these two targets fail in CI.

There's probably more errors but I cannot spot them and/or know how to fix them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ script:
       run \
       -s \
       --local_resources=400,1,1.0 \
-      //tests/examples/example_binary:hello
+      //tests/examples/example_binary:hello.exe
   - |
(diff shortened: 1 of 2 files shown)


toqueteos:
It seems Azure DevOps Pipeline does actually work, can any of you guys disable the TravisCI & AppVeyor integrations and add one that uses Azure Pipelines? /cc @tomaszstrejczek @zaphar

tomaszstrejczek:
Azure pipelines Linux and osx builds run on Microsoft hosted build agents. However, Windows build requires some specific configuration (mainly symbolic linking enabled) and it runs on my private build agent which is usually offline for cost saving.

Issue #142, opened 2019-08-10
.NET rules produce unusable .NET Standard assemblies


jimevans:
When building .NET Standard 2.0 assemblies with the current .NET rules for Bazel, the resulting assembly is not usable as a reference in .NET Core projects. The symptom of the issue is when using the resulting assembly, you receive compilation errors of the form:

    error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'.

To reproduce, clone the example project located at https://github.com/jimevans/BazelDotNetStandardExample

Execute the following command:

    bazel build //BazelDotNetStandardExample/ClassLibrary1:example

If you use the resulting assembly (which can be found at `bazel-out\x64_windows-fastbuild\bin\BazelDotNetStandardExample\ClassLibrary1\ClassLibrary1.dll`) as a reference in a .NET Core project, you will be unable to create an instance of type `Class1` from the library.

Examining the resulting assembly using ildasm, the references created by the rules are not the same as the ones created by Visual Studio for the same project. To wit, the references in the manifest for the assembly generated by the Bazel rules are:

> // Metadata version: v4.0.30319
.assembly extern System.Private.CoreLib
{
  .publickeytoken = (7C EC 85 D7 BE A7 79 8E )                         // |.....y.
  .ver 4:0:0:0
}
.assembly extern System.Private.Uri
{
  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A )                         // .?_....:
  .ver 4:0:5:0
}

By contrast, the references in the manifest for the assembly generated by Visual Studio for the same project are:

>// Metadata version: v4.0.30319
.assembly extern netstandard
{
  .publickeytoken = (CC 7B 13 FF CD 2D DD 51 )                         // .{...-.Q
  .ver 2:0:0:0
}

jimevans:
Ideally, I believe a way forward would be to use the binaries contained in the [NETStandard.Library NuGet package](https://www.nuget.org/packages/NETStandard.Library/2.0.3) as the reference assemblies that make up the stdlib for the SDK (instead of the ones that ship with the .NET Core SDK). I'm not entirely sure how to manage that, whether a new type of library generation rule (that would live alongside `dotnet_library`, `core_library`, and `net_library`) is needed, or if things can be shoehorned into working directly with one of them.

[...]

PR #143, opened 2019-08-19
Adding support for correctly building .NET Standard assemblies

jimevans:
This change introduces a new rule type, netstandard_library, and
all of the supporting data structures required to build it (context,
SDK definition, toolchain, etc.). Using the netstandard_library rule,
it should now be possible to correctly build a .NET Standard 2.0
assembly, which can be used, in turn, as a reference in a .NET Core
project.

This commit only supports creating .NET Standard 2.0 or greater
assemblies.

Fixes issue #142

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,6 +1,6 @@
 package(default_visibility = ["//visibility:public"])
 
-load("@io_bazel_rules_dotnet//dotnet/private:context.bzl", "core_context_data", "dotnet_context_data", "net_context_data")
+load("@io_bazel_rules_dotnet//dotnet/private:context.bzl", "core_context_data", "dotnet_context_data", "net_context_data", "netstandard_context_data")
 load("//dotnet:defs.bzl", "DEFAULT_DOTNET_CORE_FRAMEWORK", "DOTNET_CORE_FRAMEWORKS", "DOTNET_NET_FRAMEWORKS")
 
 dotnet_context_data(
@@ -17,6 +17,10 @@ net_context_data(
     name = "net_context_data",
     framework = "net472",
     libVersion = "4.7.2",
+)
+
+netstandard_context_data(
+    name = "netstandard_context_data",
 )
 
 exports_files(["AUTHORS"])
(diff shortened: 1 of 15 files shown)


tomaszstrejczek:
It seems that the PR doesn't build on my CI (all platforms): https://dev.azure.com/tomaszstrejczek/rules_dotnet/_build/results?buildId=347

Moreover, I think .NET Standard should be a target framework not a separate toolchain. .NET Standard assemblies may be built using classic .NET Framework toolchain, Core toolchain and Mono toolchain. 
Right now the support for .NET Standard is implemented in .NET Framework toolchain (although I haven't tested it yet). Unfortunately the implementation hardcodes using net472 version of the compiler.
Ideally the target framework should be provided as an argument to XXX_register_sdk function and  the function should download it via NuGet.

jimevans:
I don't care _**how**_ it gets solved; I care that it _**gets**_ solved. I'll see if I can get the PR to build cleanly. The tooling in the .NET Framework toolchain does not, in fact, build a .NET Standard assembly properly, at least not for me.

Honestly, I have no idea what I'm doing when creating Bazel rules, nor do I really want to. This PR is a blatant copypasta/cargo cult implementation of just trying things to get a result I can live with. I submitted it merely because the issue to which it's linked was uncommented on for a week, and my project requires that it work. If you have a better approach, I'd be happy to just have the issue fixed.

jimevans:
@tomaszstrejczek With the latest commit to the `WORKSPACE` file, the PR now builds, at least when executing `bazel build //...`. It also properly executes `bazel test //...` on my machine. However, the CI build appears to attempt to execute `bazel run //tests/examples/example_binary:hello` and `bazel run //tests/examples/example_binary:hello-core`, and these targets are not present in the build file, as near as I can tell.

tomaszstrejczek:
To somewhat limited extent .NET Standard builds are now supported. See https://github.com/bazelbuild/rules_dotnet/tree/master/tests/examples/netstandard.

PR #155, opened 2020-02-23
Make the dotnet (mono) rules work sandboxed

ozio85:
In order to make the the actions work sandboxed on linux, all files that are needed for the actions needs to be declared.

Out of all actions.run, 3 needs to be updated, the rest are ok.

Also, when the inputs are declared, the error in the runfiles action gets exposed (you cannot assume the symlink is valid to the binary action, since the toolchain is not part of the action itself).

To be safe, the binary should be copied, and not symlinked.

### DIFF
--- a/dotnet/private/actions/assembly.bzl
+++ b/dotnet/private/actions/assembly.bzl
@@ -146,7 +146,7 @@ def emit_assembly(
 
     deps_files = [d[DotnetLibrary].result for d in transitive.to_list()]
     dotnet.actions.run(
-        inputs = attr_srcs + [paramfile] + deps_files + [dotnet.stdlib] + [r[DotnetResource].result for r in resources],
+        inputs = attr_srcs + [paramfile] + deps_files + [dotnet.stdlib, dotnet.mcs] + [r[DotnetResource].result for r in resources],
         outputs = [result] + ([pdb] if pdb else []),
         executable = dotnet.runner,
         arguments = [dotnet.mcs.path, "/noconfig", "@" + paramfile.path],
--- a/dotnet/private/actions/assembly_core.bzl
+++ b/dotnet/private/actions/assembly_core.bzl
@@ -144,7 +144,7 @@ def emit_assembly_core(
 
     deps_files = [d[DotnetLibrary].result for d in transitive.to_list()]
     dotnet.actions.run(
-        inputs = attr_srcs + [paramfile] + deps_files + [dotnet.stdlib] + [r[DotnetResource].result for r in resources],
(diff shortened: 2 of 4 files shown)


ozio85:
@googlebot I signed it!

tomaszstrejczek:
Unfortunately the patch fails under linux and osx:
https://dev.azure.com/tomaszstrejczek/rules_dotnet/_build/results?buildId=449&view=results
The error message is the following: 
ERROR: /home/vsts/work/1/s/tests/examples/example_tool/BUILD:42:1: declared output 'tests/examples/example_tool/hello_world.txt' was not created by genrule. This is probably because the genrule actually didn't create this output, or because the output was a directory and the genrule was run remotely (note that only the contents of declared file outputs are copied from genrules run remotely)
ERROR: /home/vsts/work/1/s/tests/examples/example_tool/BUILD:42:1: not all outputs were created or valid

ozio85:
I have been looking into the problem.. but i am afraid that the problem might be slightly deeper than just changing to "copy". I need to create a launcher from the toolchain instead, and let the launcher be located in dotnet_sdk. There is a skylib rule that can create a quite elegant launcher, and then we could remove much of the dotnet/tools/ logic..

Please advice.. this might be a slightly bigger change, but might be worth considering if we can get remote execution and sandboxing to work

tomaszstrejczek:
Sounds great. Dotnet/tools logic is quite shaky now and it would be great to simplify it. Getting sandboxing and remote execution to work would make it even better.

ozio85:
This is the progress so far. I have something that works, sandboxed and remote.. but it requires an additional "runner" to the binary:

So upsides:
* Works sandboxed and remote
* No copying of runtime dependencies (no need for _copy or _symlink), we can use the standard rlocation strategy.
* No need of the launcher (the existing sh_binary is used, which creates the exe for us)
* No need to copy/symlink the dotnet runner to the run-location
* No need to create a runner subfolder

Downsides:
* The binary itself is not runnable with "run", you have to run the <binary>_runner target :/

I will look into other projects how they have solved it .. but I see that this solution breaks your strive to keep it simple for the user.

```
def _binary(kind, name, **kwargs):
    kind(
        name = name,
        **kwargs,
    )

    sh_binary_with_rlocation(
        name = name + "_runner",
        srcs = [":" + name],
    )

def dotnet_binary(name, **kwargs):
    _binary(_dotnet_binary, name, **kwargs)

def core_binary(name, **kwargs):
    _binary(_core_binary, name, **kwargs)

def net_binary(name, **kwargs):
    _binary(_net_binary, name, **kwargs)
```

tomaszstrejczek:
Closing  due to inactivity. Superseded by #186.

PR #156, opened 2020-02-26
Define Windows flag if it is not available

plule-ansys:
The `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` was introduced fairly recently in the Windows SDK, making it unavailable in some environments.

This PR defines it if not available. It is defined to its actual value, so if the underlying Windows version supports it but the Windows SDK is outdated, the feature will still work.

The same workaround is available in [bazel itself](https://github.com/bazelbuild/bazel/blob/master/src/main/tools/build-runfiles-windows.cc#L39)

### DIFF
--- a/dotnet/tools/common/manifest.c
+++ b/dotnet/tools/common/manifest.c
@@ -11,6 +11,9 @@
 #include <process.h>
 #define F_OK 0
 #pragma comment(lib, "shlwapi.lib")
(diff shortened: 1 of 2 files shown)


plule-ansys:
I will sort out the cla issue as soon as possible

Issue #151, opened 2019-11-28
Referencing AspNetCore dlls in .NET Core 3.0.100 SDK


ztl8702:
It seems that AspNetCore 3.0 is [no longer shipped as nuget packages](https://stackoverflow.com/a/58126758) but rather inside the SDK. 

i.e. it is located in `shared/Microsoft.AspNetCore.App/3.0.0`.

To reference it, I assume we need to add `core_stdlib` rules for Microsoft.AspNetCore.App DLLs to  https://github.com/bazelbuild/rules_dotnet/blob/master/dotnet/stdlib.core/macro.bzl ?

@tomaszstrejczek  can you show me how you created [those rules](https://github.com/bazelbuild/rules_dotnet/blob/eaa368a35b6d0629a240ba6422f210e9dec1c241/dotnet/stdlib.core/macro.bzl#L9-L174) for Microsoft.NETCore.App DLLs (and their deps)? Was that by hand or using some tools? I would love to help.

codebreach:
any way I can help with this? we also need this support!

tomaszstrejczek:
The support for ASP.NET Core build is now available on master. See https://github.com/bazelbuild/rules_dotnet/tree/master/tests/examples/aspnetcore

ztl8702:
Cool. Thanks!

PR #157, opened 2020-03-26
ASPNetCore Support for 3.x

codebreach:
Fixes https://github.com/bazelbuild/rules_dotnet/issues/151

Please note that this is a large change and I am all ears if you would rather have it implemented in a different way.

Add a new tool shared2bazel that creates dependencies for stdlib
This tool uses Reflection to figure out which Assembly References exist

Added `all_core_stdlib_above3` to macro.bzl which exposes new dlls from stdlib
`all_core_stdlib_above3` also expsoses all ASPNETCORE DLLs

Added `extra_shared` to context.
`extra_shared` is used by `_stdlib_impl` and `_get_shared_dir` to get the right path.

Updated CopyRunfiles to not copy `.version` twice.

Updated core-3.0.100 and core-3.1.100 to expose new DLLs

Updated `DOTNET_CORE_FRAMEWORKS` to include shared frameworks


codebreach:
@googlebot I signed it!

tomaszstrejczek:
Since CLA is not signed I've implemented ASP.NET Core support in a different way. See https://github.com/bazelbuild/rules_dotnet/tree/master/tests/examples/aspnetcore.

codebreach:
Wow that's a much cleaner implementation aswell. I'll try it out on our
repo and let you know if there are any issues.

Thanks for the update!

PR #195, opened 2020-08-11
Fix issues with sync command

J-Swift:
There was a mismatch between what looks like old default values and new default values. Also the configuration passed was not being used.


J-Swift:
@googlebot I signed it!

PR #212, opened 2021-02-06
.Net 5 Support

### LINKED ISSUE #210: .Net 5 Support
Would it be possible to add .Net 5 support? I tried looking into it but I did not figure out how to generate the generate.bzl and generate2.bzl files that are there for each version.

purkhusid:
## What?
Adds .Net 5.0 support
Changes the `nuget2bazel/rules` command so that it can be run on Linux/MacOS
Makes .Net 5.0 the default SDK

## How?
Had to make a few changes to the `nuget2bazel/rules` command
so that I could run it on Linux. Afterward I ran the command after
adding the new SDK to the SDK list to generate the core library targets

Closes #210


purkhusid:
Superseded by #240

PR #215, opened 2021-02-08
F# Support

purkhusid:
This is a WIP PR of F# support

It seems like it's working but I would like to port the examples to F# to have better tests

### DIFF
--- a/dotnet/private/actions/assembly_common.bzl
+++ b/dotnet/private/actions/assembly_common.bzl
@@ -51,8 +51,31 @@ def emit_assembly_common(
       target_framework: target framework to define via System.Runtime.Versioning.TargetFramework
       nowarn: list of warnings to ignore
       langversion: version of the language to use (see https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version)
+      version: the version of the generated assembly
     """
+    if dotnet.lang == "csharp":
+        return _emit_assembly_common_csharp(kind, dotnet, name, srcs, deps, out, resources, executable, defines, unsafe, data, keyfile, subdir, target_framework, nowarn, langversion, version)
+    else:
+        return _emit_assembly_common_fsharp(kind, dotnet, name, srcs, deps, out, resources, executable, defines, unsafe, data, keyfile, subdir, target_framework, nowarn, langversion, version)
 
+def _emit_assembly_common_csharp(
+        kind,
+        dotnet,
+        name,
+        srcs,
+        deps = None,
+        out = None,
+        resources = None,
+        executable = True,
+        defines = None,
+        unsafe = False,
+        data = None,
+        keyfile = None,
+        subdir = "./",
+        target_framework = "",
+        nowarn = None,
+        langversion = "latest",
+        version = (0, 0, 0, 0, "")):
     if name == "" and out == None:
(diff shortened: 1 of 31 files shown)


purkhusid:
@zaphar @tomaszstrejczek I would love some input on the approach I'm taking and if you feel like this approach is a good one.

dotnet/private/actions/assembly_common.bzl
```diff
@@ -218,3 +243,184 @@ def emit_assembly_common(
...
+    args.add("/highentropyva+")
+    args.add("/noframework")
+    args.add("/nowin32manifest")
+    args.add("/targetprofile:netcore")
+    args.add("/simpleresolution")
+    args.add("/nocopyfsharpcore")
+
+    # Debug parameters
+    if pdb:
+        args.add("-debug:full")
```
baronfel, line 303: `full` isn't useful for .net core anymore, all tools should be using `portable` (if they want standalone PDBs), or `embedded` (if they want PDBs embedded in the output assembly itself).

baronfel: or you could do like the C# targets a bit above and put `debug:full` behind a check for `mono`, but I wouldn't even suggest that, as mono can read the `portable` pdb format.

tomaszstrejczek: I would consider supporting F# via separate toolchain. The version of rules_dotnet I'm working on is moving towards deeper support of Bazel platforms & toolchains.

purkhusid:
> I would consider supporting F# via separate toolchain. The version of rules_dotnet I'm working on is moving towards deeper support of Bazel platforms & toolchains.

@tomaszstrejczek Are there some major refactors coming up? If you are working on some major refactor I can postpone this and reintroduce it after you are done.

Also, if you are going to do work on the toolchains then it would be nice if it could be done in a way that makes it easy to share between .Net languages. E.g. in this PR you can see that I could use almost all of the work you had done with C# except the compilation step.

tomaszstrejczek:
@purkhusid Yes, I've just merged significant changes. 

I think that support for new language should be done in the following way:
1. New toolchain_type should be added to BUILD.
2. New toolchains should be defined in declare_toolchains(). New toolchains may reuse macro core_toolchain, but an attribute must be added to specify the language/compiler and toolchain type. 
3. New toolchain may reuse emit_assembly_core. 
3. New rules have to be defined for F# source code. The rules should use new toolchain_type and reuse current implementation functions.

purkhusid:
> @purkhusid Yes, I've just merged significant changes.
> 
> I think that support for new language should be done in the following way:
> 
>     1. New toolchain_type should be added to BUILD.
> 
>     2. New toolchains should be defined in declare_toolchains(). New toolchains may reuse macro core_toolchain, but an attribute must be added to specify the language/compiler and toolchain type.
> 
>     3. New toolchain may reuse emit_assembly_core.
> 
>     4. New rules have to be defined for F# source code. The rules should use new toolchain_type and reuse current implementation functions.

Ok, I'l take another look at this. Would you be up for changing the names of the current rules from `core_library/core_binary` to `csharp_library/csharp_binary` ?

tomaszstrejczek:
Superseded by #217

PR #217, opened 2021-02-11
F# Support - Take 2

purkhusid:
## What?
This is another iteration of adding F# support after a recent toolchain refactor.

## How?
* Renamed the C# toolchain
* Added a new F# toolchain
* Renamed `core_library`, `core_binary`, `core_xunit_test` and `core_nunit3_test` to
`csharp_library`, `csharp_binary`, `csharp_xunit_test` and `csharp_nunit3_test`
* Added new rules `fsharp_library`, `fsharp_binary`, `fsharp_xunit_test` and `fsharp_nunit3_test`
* Added tests using the F# rules


purkhusid:
@tomaszstrejczek What do you think of this approach?

The only thing that I think might be unecessary is having a separate toolchain for C# and F# since they are almost exactly the same. It complicates things a bit regarding rules that are not language specific, like `core_resource`. 

If we go the way of having a separate C# and F# toolchain I think it might be best to create a 3rd toolchain for just the runtime/stdlib.

Othewise the abstraction in the rules made it very easy to add F# support and I'm looking forward to your feedback!

tomaszstrejczek:
@purkhusid Looks great. Thank you.

PR #231, opened 2021-02-28
Add support for extracting nuget packages from project files

thelgevold:
Proposing an addition to the Nuget2Bazel tool that will allow us to extract nuget packages from VS project files. 

Not sure if the command line interface is ideal, but I have added a new option (-f) as seen below:

```
bazel run //tools/nuget2bazel:nuget2bazel.exe -- add -p [WORKSPACE path] -f project1.csproj subfolder/project2.csproj
```
The list of project files is space delimited, and any overlapping packages will be de-duped.

Thoughts?


tomaszstrejczek:
Looks great. Thank you.

PR #232, opened 2021-03-06
Fix F# type providers

purkhusid:
Attemtpting to fix F# type providers


purkhusid:
This should be ready for review. I did not implement transitive passing of design time resources so users types that are generated with Type Providers will also have to supply the design time resources in their own target.

This could probably be changed to pass the transitive design time resources but I did not implement that.
If anybody has suggestions on how that might be best done then that is appreciated.

PR #273, opened 2022-01-25
Add .NET SDK 6.0.101. Darwin ARM64 support in nuget2bazel.

mjs-rokt:
- Support Darwin ARM64 support in `nuget2bazel`.
- Adds .NET SDK 6.0.101.
- Removes use of `alias` in generated runtime Starlark files specifically for .NET 6, where this does not appear to be necessary any longer.
- Updates some usages of `@bazel_tools` to `@platforms`.
- Adds .NET 6.0.101 to the presubmit.

Currently `nuget2bazel` assumes `amd64` everywhere. The implementation here relies on conditionals to keep the diff size down at the cost of complexity.

Eventually `nuget2bazel` should be properly refactored to support more architectures (i.e. Linux ARM64), but that change might result in a much larger diff.

This PR opts for having .NET 6 support today with a smaller but less clean patch.


mjs-rokt:
Looks like CI is failing with: `The number of tasks in one config file is limited to 50!` (https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/bazelci.py#L998).

Hopefully a maintainer can chime in and suggest what to do about the tests.

purkhusid:
@mjs-rokt I removed a bunch of tasks from the presubmit.yml. For now we'll just run CI for the latest SDK version for each channel.

tools/nuget2bazel/rules/RuntimeCoreGenerator.cs
```diff
@@ -44,7 +44,7 @@ private async Task Handle(StreamWriter f, Sdk sdk)
             await f.WriteLineAsync("def define_runtime():");
             await f.WriteLineAsync("    \"Declares runtime\"");
 
-            if (sdk.Packs != null)
+            if (sdk.Packs != null && !sdk.Version.StartsWith("6"))
```
purkhusid, line 47: Why is this change needed?

purkhusid:
The CI fails now because you'll also have to add `6.0` here: https://github.com/bazelbuild/rules_dotnet/blob/e081c2a989999a47126bcb5bc5699777f3714c5a/tools/nuget2bazel/WorkspaceEntry.cs#L32 and then regenerate this file https://github.com/bazelbuild/rules_dotnet/blob/78f9b2a243175c54686a5f7c9ca3f060c133175e/dotnet/private/deps/nuget.bzl#L19 using `nuget2bazel`

mjs-rokt:
I added a dependency update script for `bash` called `deps.sh` and ran it to add `net6.0` nuget references. The script removes two entries from the list in `all.bat`:

- NuGet.Packaging
- NuGet.Protocol

Since these resulted in `nuget2bazel` errors like: `'NuGet.Packaging.Core.5.11.0' already exists in folder`.

tools/nuget2bazel/rules/RuntimeCoreGenerator.cs
mjs-rokt: The aliases created here already appear as generated `core_stdlib_internal` target names in `stdlib/6.0.101.bzl` after running `nuget2bazel`.

There are no actual targets prefixed with `p1_...` to reference with the alias in the 6.0 generated code, since these DLLs are generated directly with the expected name: https://github.com/bazelbuild/rules_dotnet/pull/273/files#diff-82c7e5d7038182e73763f0843832c48dde3e3c0a374f0e1432d51d63964cf97fR1585.

This change removes the aliases since the `p1_` prefix does not exist. This only happens for code generated for the 6.0 runtime, and not for previous runtimes, where the `p1_` prefixes still exist.

mjs-rokt:
What should we do about `arm64` tests? Looks like CI does not have a configuration for that machine type: https://buildkite.com/bazel/rules-dotnet-edge/builds/1066#f0702114-c8c0-4a1e-901d-cb8b2d41453d/247-294

mjs-rokt:
@purkhusid Finally have CI passing with a couple of changes:

1. Removing the `arm64` tests since there are no runners (described above)
2. Patching `rules_proto_grpc` to include `net6.0` references in its nuget packages.

The patch included in this PR is just to verify that the CI passes - should we remove the patch, mark the test as incomaptible with net6.0, and then reintroduce the test for 6.0 once rules_proto_grpc is updated?

purkhusid:
> The rules_proto_grpc patch included in this PR is just to verify that the grpc test CI passes - should we remove the patch, mark the test as incompatible with net6.0, and then reintroduce the test for 6.0 once rules_proto_gr

That sounds good to me!

mjs-rokt:
@purkhusid I've removed the patch and CI is passing - let me know if there's anything else I should address.

purkhusid (approved):
This looks great! If you have the time to patch the upstream proto-grpc rules then that would be very nice!

Thanks for the contribution!

PR #285, opened 2022-05-16
Next: Configurable toolchains

purkhusid:
This change makes the dotnet SDK configurable with a repository rule.
You can select the SDK version and the available SDKs can be easily updated by running `dotnet fsi ./scripts/UpdateSdks.fsx`

Included in this is MacOS ARM support. 

I also decided that the minimum supported SDK version would be 6.0 since MacOS ARM support came in 6.0.


purkhusid:
This PR closely follows the pattern used here: https://github.com/bazel-contrib/rules-template

PR #292, opened 2022-06-21
Add a runfiles library

purkhusid:
## What?
* Adds a runfiles lib for .Net
* Adds `data` attribute to all rules
* Adds an example using the runfiles lib

## How?
The .Net runfiles lib is an translation of the Java runfiles library that can be found here: https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java

## Why?
Using `data` for various runtime files is very idiomatic when using Bazel and having a runfiles library takes care of all the complicated stuff around Bazel runfiles.


tomdegoede:
@purkhusid looks pretty good. We had a custom thing for this but since this one reflects a Java variant I think it's better. There are some minor optimizations there could be made such as just reading env vars directly instead of using a dict but nothing significant to block the PR.

PR #301, opened 2022-08-05
Dependency resolution fixes

### LINKED ISSUE #297: [Next] Expecto cannot find method implementation from a dependency
Probably easiest to just see this repo: https://github.com/njlr/bazel-expecto-repro

Basically I have a `netstandard2.1` library that I am testing using a `net6.0` binary. 

The code builds with Bazel and the tests pass with MS Build.

```bash
$ dotnet run --project ./ratio.tests
[20:34:49 INF] EXPECTO? Running tests... <Expecto>
[20:34:49 INF] EXPECTO! 1 tests run in 00:00:00.0524229 for Ratio.+ works for simple cases – 1 passed, 0 ignored, 0 failed, 0 errored. Success! <Expecto>
```

However, running the test (`bazel test //...`) with Bazel gives a run-time error with Expecto tests:

```
System.TypeLoadException: Method 'Equals' in type 'Ratio.Ratio' from assembly 'ratio, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.
   at Program.tests@10-1.Invoke(Unit unitVar)
   at Expecto.Impl.execTestAsync@569-1.Invoke(Unit unitVar) in /Users/h/dev/haf/expecto/Expecto/Expecto.Impl.fs:line 576
   at Microsoft.FSharp.Control.AsyncPrimitives.CallThenInvoke[T,TResult](AsyncActivation`1 ctxt, TResult result1, FSharpFunc`2 part2) in D:\a\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 447
   at Microsoft.FSharp.Control.Trampoline.Execute(FSharpFunc`2 firstAction) in D:\a\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 104
```

Maybe I'm just missing something simple though?

purkhusid:
## What?
* Adds an incoming transition to _binary/_library rules that transitions to the highest compatible framework in the `target_frameworks`
* Adds a `private_deps` attribute to _binary/_library rules that is intended for compile time dependencies. These dependencies are not transitively passed and do not become part of the runfiles.
* Some cleanup and reorganizing

## Why?
The transition is needed because before the change we were just taking the default tfm and passing it to the build actions. This did not make sense since you could update rules_dotnet and then the default tfm would be 6.0 but your targets would still target 5.0. There was also an issue that if you were e.g. targeting `netstandard2.1` we would still pass the default tfm to the build actions. 

The `private_deps` attribute is needed so that we can pass in targeting packs without passing them transitively to parents. Targeting packs should not be passed transitively because that causes conflicts between targeting packs. 

Fixes #297


purkhusid:
@njlr I tested this on the repro you sent in #297 and this fixes it.

PR #303, opened 2022-08-19
Publishing and native dependencies

purkhusid:
This PR got a bit bloated since I used it to get the
`next` branch to a workable state within the monorepo
that my work has set up.

## What?
The big featues in this PR are:
* Binary publishing with support for runtime identifiers
* Support for the `runtimes` folder in NuGet folders 
* Add support for configuring compiler warnings/errors globally
or on per target basis

Various small tasks:
* Improve the Runfiles library so that it works with publishing
* Removed dependency on app host NuGet since it's no longer
updated and now we use the apphost DLL that is part of the SDK
* Update CI scripts to run all tests and examples
* Update some READMEs to include up to date information
    * CONTRIBUTING.md
    * Root README.md
    * Deleted docs that were no longer relevant
* Add auto generation of documentation
    * The current state of the docs is not ready though.
* Add pre-commit config that checks that everything is formatted
* Update gitpod setup so that it install pre-commit hooks
* Split up rule dependencies so that they can be declared more granularly
* Update Gitpod setup


.bazelrc
```diff
@@ -3,6 +3,9 @@ startup --windows_enable_symlinks
 build --enable_runfiles
 test --test_output=errors
 
+# So that the bazel-bin folder is populated since we use transitions heavily
+--use_top_level_targets_for_symlinks
```
tomdegoede, line 7: Is this recommended for any rules_dotnet bazel repo? I don't really understand the reason yet.

dotnet/private/rules/publish_binary/publish_binary.bzl
```diff
@@ -0,0 +1,376 @@
...
+    # DefaultInfo provider and can thus be easily forwarded to filegroups/tars/containers.
+    #
+    # The runfiles library follows the spec and tries to find a `<DLL>.runfiles` directory
+    # next to the the DLL based on argv0 of the running process if
+    # RUNFILES_DIR/RUNFILES_MANIFEST_FILE/RUNFILES_MANIFEST_ONLY is not set).
+    for file in assembly_info.data + assembly_info.transitive_data.to_list():
+        inputs.append(file)
+        manifest_path = _to_manifest_path(ctx, file)
+        output = ctx.actions.declare_file(
+            "{}/publish/{}/{}.runfiles/{}".format(ctx.label.name, runtime_identifier, binary_info.app_host.basename, manifest_path),
```
tomdegoede, line 153: This looks quite scary. Are you reconstructing the runfiles folder. Doesnt this collide with the native bazel runfiles symlink tree? I think we should try to match msbuild publish behavior and just copy all binaries to the root publish folder except for runtime specific ones.

dotnet/private/tests/nuget_structure/common.bzl
```diff
@@ -1,37 +1,40 @@
...
-    prefs_files = _get_nuget_relative_paths(provider.prefs)
+    prefs_files = _get_nuget_relative_paths(provider.ref)
     asserts.true(
         env,
         sorted(ctx.attr.expected_refs) == sorted(prefs_files),
         "\nExpected prefs:\n{}\nActual prefs:\n{}".format(ctx.attr.expected_refs, prefs_files),
     )
 
-    irefs_files = _get_nuget_relative_paths(provider.irefs)
+    irefs_files = _get_nuget_relative_paths(provider.iref)
```
tomdegoede, line 37: Not sure why these renames were needed or make sense.

dotnet/private/rules/publish_binary/publish_binary.bzl
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+    ctx.actions.write(
+        output = output,
+        content = json.encode_indent(base),
+    )
+
+# TODO: Reuse this when we create the deps.json in the csharp_binary/fsharp_binary rules
+# For deps.json spec see: https://github.com/dotnet/sdk/blob/main/documentation/specs/runtime-configuration-file.md
+def _generate_depsjson(
```
tomdegoede, line 218: nice!

tomdegoede (approved):
I have reviewed some changes but was not able to look at each file individually. I'm getting a bit worried that the code base is getting incomprehensibly big. We should try to minimize the individual changes in the future, and separate refactoring/chore commits from logical ones. 

Could we try to group certain aspects of the implementation in a more modular fashion at the root level? Currently most logic is inside `dotnet/private/rules` but we could try to split dotnet into `nuget` for dependency resolution, `roslyn` for compilation and `dotnet` for runtime specific logic. What do you think?

dotnet/private/rules/publish_binary/publish_binary.bzl
tomdegoede: I just ran the example but don't see the .runfiles folder in the output. Can you explain why it's mentioned here?

tomdegoede: O I see the pdb file is written there.

purkhusid: The reason or this is that runfiles have generally had a bad track record with packaging rules
See https://github.com/bazelbuild/rules_pkg/pull/605 for some discussion. 

The only files that are put into the runfiles folder are the items that go into `data` on each rule. This is so that the runfiles library can find the data files under the expected location.
All other files follow the standard .Net convention where everything is copied to the root of the publish folder.

I agree that this isn't very orthodox but this makes it possible to publish  the applications without having to wrangle possible runfile issues when you e.g. want to zip/tar/containerize the output of this rule. I can try using the standard runfiles once the runfiles support lands in rules_pkg.

purkhusid: Regarding the PDB files, I was thinking about adding a specific attribute for them in the providers and not add them to `data`. That way they would just also end up in the root publishing folder.

dotnet/private/tests/nuget_structure/common.bzl
purkhusid: Yeah, I originally did them because they reflect the folders in the nuget packages. But I'll change it back I think.

dotnet/private/rules/publish_binary/publish_binary.bzl
tomdegoede: Hmm I dont think I like this approach. We want to be as close as possible to msbuild logic. So when we have some assets flagged as copy to output, they end up in data[]. They should be copied to the publish folder without the runfiles subdir since this is not a msbuild concept. I'm not sure how msbuild handles directory structures here, but we should try to copy that behavior.

purkhusid: The `data` attribute is very Bazel specific so I don't think we should conflate it with the `copy to output` functionality of MSBuild. If we want to add that functionality we should add another `copy_to_output` attribute that e.g. takes in a map of output pathts and file labels.

We don't want to break the Bazel conventions as well which is very well defined for the `data` attribute.

tomdegoede: hmm you mean the magic propagation of data attribute values? I'm not quite sure about it's use case trough bazel conventions but I have read about it. We currently use data[] to propagate the copy to output files, for example for assets that used during tests.

tomdegoede: A separate copy_to_output sounds reasonable

purkhusid: Yeah, I agree that the `data` attribute can be quite obscure to those who are not used to Bazel but I don't think we should break the set convention where the `data` files should be part of the runfiles.

We could also add an attribute `copy_data_to_outputs` which copies the data files to the output folder while keeping the relative folder structure.

purkhusid:
@tomdegoede I cleaned up a bit in the folders and I think it's pretty good so far. Having the `dotnet` folder in the root of the repo is following the guidelines posted here: https://docs.bazel.build/versions/main/skylark/deploying.html
So we have `dotnet/private/rules/{common, csharp, fsharp, nuget, publish}`

PR #316, opened 2022-11-12
next-assembly-info-files

njlr:
This PR adds a new attribute to the `DotnetAssemblyInfo` provider called `files`, which is all of the files in the package. This is ground-work for building a `rules_fable`.

Fable will attempt to "crack" a Nuget package by starting from a referenced DLL and walking up the directory tree till a `nuspec` file is found. Then it will try to collect any `fs`, `fsi` and `fsproj` files in that sub-tree. To support Fable without major modifications, it is easiest to give all package files as inputs to the Bazel action. 

Details here: https://github.com/fable-compiler/Fable/blob/main/src/Fable.Cli/ProjectCracker.fs


njlr:
https://github.com/bazel-contrib/rules_dotnet/pull/455

PR #319, opened 2022-12-01
add support for .net 7.0.100

s3w3nofficial:
add support for .net 7.0.100


purkhusid:
There is now support for .Net 7 on `master` after the `next` branch was merged.

PR #335, opened 2023-02-04
fixed nuget structure for netstandard20 and netstandard21

Place1:
I noticed that some nuget packages target `netstandard21` and `netstandard20` instead of `netstandard2.1` and `netstandard2.0`. This leads to nuget_archive not generating the "libs" select correctly.

This PR updates the nuget_archive logic to fix this issue.

I also added a test that checks the generated structure for `Magick.NET.Core` [source](https://github.com/dlemstra/Magick.NET/blob/main/src/Magick.NET.Core/Magick.NET.Core.csproj#L3)


purkhusid (approved):
Looks good, thanks for the contribution!

Issue #298, opened 2022-07-21
Question: is there a way to use the `dotnet` executable


njlr:
The context is to be able to execute pre-built assemblies with `dotnet`. 

I noticed a substitution variable here:

https://github.com/bazelbuild/rules_dotnet/blob/5c95c346d2b362d56eeada72bdf9d42395eae405/dotnet/toolchain.bzl#L91

Is there an example of how to use it, e.g. in a `genrule`?

njlr:
Some progress...

```starlark
genrule(
  name = "test",
  outs = [ "out.txt" ],
  cmd = "echo \"$(DOTNET_BIN)\" > $@",
  toolchains = [
    "@dotnet_toolchains//:resolved_toolchain",
  ],
)

```

```bash
cat ./bazel-bin/out.txt 
external/dotnet_x86_64-unknown-linux-gnu/dotnet
```

tomdegoede:
We do something similar in our private repository by just invoking `bazel run @core_sdk//:dotnet.exe` commands. I'm not sure where it'll be located for you but I'm sure you should be able to do the same. Probably there will be an external repo for your current platform such as `dotnet_x86_64-pc-windows-msvc`, but we could simply expose a .sh and .bat file to proxy it via a more generic name.

tomdegoede:
Do you want to use dotnet.exe from within a bazel action or just from the commandline? From your genrule example I think the latter, in which case you could look into importing the prebuilt binaries as a rules_dotnet executable.

tomdegoede:
I think we are missing an import_binary rule here https://github.com/bazelbuild/rules_dotnet/blob/next/dotnet/private/rules/imports.bzl

purkhusid:
> ```python
> toolchains = [
>     "@dotnet_toolchains//:resolved_toolchain",
>   ],
> ```

This should work but you also need to add
```
tools= [
    "@dotnet_toolchains//:resolved_toolchain",
  ],
```

purkhusid:
> Some progress...
> 
> ```python
> genrule(
>   name = "test",
>   outs = [ "out.txt" ],
>   cmd = "echo \"$(DOTNET_BIN)\" > $@",
>   toolchains = [
>     "@dotnet_toolchains//:resolved_toolchain",
>   ],
> )
> ```
> 
> ```shell
> cat ./bazel-bin/out.txt 
> external/dotnet_x86_64-unknown-linux-gnu/dotnet
> ```

[...]

PR #336, opened 2023-02-06
fix: include runtime files in resolved toolchain #298

Place1:
This PR updates the `@dotnet_toolchains//:resolved_toolchain` target to include the dotnet runtime files so that it's possible to execute the `dotnet` binary from a genrule.

I've also added a testcase that executes `dotnet --version` from a genrule and then asserts the output. I'm a complete bazel novice (been using for a week) so please help me if this test isn't following bestpractice :)


Place1:
Hmm, the build fails in RBE because of

```
No .NET SDKs were found.
```

purkhusid:
@Place1 I took the liberty to fix this for all the tools that are in the toolchain. Thanks for the contribution!

purkhusid (approved):

Place1:
Thanks so much for taking the time to look it over an do it up! much appreciated.

PR #350, opened 2023-04-17
Disconnected TFM graphs in non deps attributes

purkhusid:
## What?
This adds an transition to all label attributes that do not already  have the TFM transition.
All this transition does it transition the TFM/RID back to the default configuration.

## Why?
Because the end user might have labels that are *_binary targets in those attributes but they should not be connected to the actual TFM graph of the build for the owning target.

Examples are:
Target A uses generated  in it's `srcs` attribute  coming from  target B
Target A uses a csharp_binary in it's data dependency because it needs to run the binary


purkhusid:
@jimevans @simonstewart This might unblock your work migrating Selenium to rules_dotnet. Could you test this out?

PR #355, opened 2023-05-13
Add attribute to set /nullable flag to C# compiler

fedemartinezdev:
Adds the attribute `nullable` to csharp. It enables nullable context, or nullable warnings. The supported values are the same as description in the [C# Compiler options](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/language#nullable), these being: `enable`, `disable`, `warnings`, or `annotations`.

### DIFF
--- a/docs/csharp_binary.md
+++ b/docs/csharp_binary.md
@@ -10,12 +10,12 @@ Rule for compiling C# binaries.
 
 <pre>
 csharp_binary(<a href="#csharp_binary-name">name</a>, <a href="#csharp_binary-additionalfiles">additionalfiles</a>, <a href="#csharp_binary-allow_unsafe_blocks">allow_unsafe_blocks</a>, <a href="#csharp_binary-apphost_shimmer">apphost_shimmer</a>, <a href="#csharp_binary-compile_data">compile_data</a>, <a href="#csharp_binary-data">data</a>,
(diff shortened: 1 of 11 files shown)


fedemartinezdev:
I am not sure why am I getting the following error, I don't think it was introduced by me but please let me know if I did some error.

```
ERROR: C:/b/bk-windows-3md3/bazel/rules-dotnet-edge/dotnet/private/tests/dependency_resolution/diamond_dependencies/a_with_only_transitive_d/BUILD.bazel:6:18: every rule of type csharp_test implicitly depends upon the target '//dotnet/private/rules/common:disable', but this target could not be found because of: no such target '//dotnet/private/rules/common:disable': target 'disable' not declared in package 'dotnet/private/rules/common' defined by C:/b/bk-windows-3md3/bazel/rules-dotnet-edge/dotnet/private/rules/common/BUILD.bazel (Tip: use `query "//dotnet/private/rules/common:*"` to see all the targets in that package)
```

fedemartinezdev:
> I am not sure why am I getting the following error, I don't think it was introduced by me but please let me know if I did some error.
> 
> ```
> ERROR: C:/b/bk-windows-3md3/bazel/rules-dotnet-edge/dotnet/private/tests/dependency_resolution/diamond_dependencies/a_with_only_transitive_d/BUILD.bazel:6:18: every rule of type csharp_test implicitly depends upon the target '//dotnet/private/rules/common:disable', but this target could not be found because of: no such target '//dotnet/private/rules/common:disable': target 'disable' not declared in package 'dotnet/private/rules/common' defined by C:/b/bk-windows-3md3/bazel/rules-dotnet-edge/dotnet/private/rules/common/BUILD.bazel (Tip: use `query "//dotnet/private/rules/common:*"` to see all the targets in that package)
> ```

Nevermind, I found my error, I was using a `label` type for the nullable attribute instead of a `string`.

fedemartinezdev:
@purkhusid do you mind reviewing this PR?

purkhusid (approved):

Issue #361, opened 2023-07-17
Is private nuget repo supported?


zhongyi-zhang:
If I follow the paket doc https://github.com/fsprojects/Paket/blob/master/docs/content/nuget-dependencies.md to use the env var approach, does the bazel generated by paket2bazel read the env vars?

Appreciate your help on this, thanks!

zhongyi-zhang:
It seems the nuget repo is even hardcoded: https://github.com/bazelbuild/rules_dotnet/blob/fa1e285adf34fbf4fc59bbfc65b6d33795b4cc67/dotnet/private/rules/nuget/nuget_archive.bzl#L280

PR #362, opened 2023-07-25
Support using custom NuGet repo and basic auth for it

zhongyi-zhang:
Solving #361.
This PR is to support using custom NuGet repo. And for private NuGet repo, we can use a netrc file which is the bazel supported auth style for a repository rule: https://bazel.build/rules/lib/repo/utils#read_netrc.

Some more notes:

* This PR only implements single NuGet repo for a `nuget_repo` target, which is a little different from the Paket usage which can set multiple resources (as repos). But I think that's acceptable as we know which repo to get the packages in most cases.

* Only basic auth is implemented in this PR. We can expose the auth type and implement other types of auth later.

* The `resource` parameter is for a package url template. It is normal for v2 repo but a little strange for v3 repo. There is an option: we can distinguish v2 or v3 repo. For v3 repo, we fetch the basic package url from the index. Not sure if it's worth.

If the change looks good to you, I am happy to update the responding documents as well.


purkhusid:
I'll take a look next week, on vacation now.

purkhusid:
I have not forgotten about this. Will take a look this week.

purkhusid:
@zhongyi-zhang I did some changes to your approach and added them to this branch. Could you try this out with your private registry and see if everything works?

dotnet/private/rules/nuget/nuget_archive.bzl
```diff
@@ -276,15 +282,64 @@ def _process_key_and_file(groups, key, file):
...
+    return package_urls
+
+def _get_auth_dict(ctx, netrc, urls):
+    # Default to the user's netrc
+    netrc = read_user_netrc(ctx)
+
+    # If there is an netrc file declared for the specific package
+    # we use that one instead of the user netrc
+    if ctx.attr.netrc:
+        netrc = read_netrc(ctx, netrc)
```
zhongyi-zhang, line 325: It should be `netrc = read_netrc(ctx, ctx.attr.netrc)`, right?

tools/paket2bazel/Gen.fs
```diff
@@ -40,8 +40,17 @@ let generateTarget (group: Group) (repoName: string) =
...
+            |> Seq.fold (fun state current -> state + $"\"{current}\", ") ""
+            |> (fun s ->
+                if String.IsNullOrEmpty(s) then
+                    s
+                else
+                    s.Substring(0, s.Length - 2))
+
         sb.Append(
-            $"{i}        (\"{package.name}\", \"{package.version}\", \"{package.sha512sri}\", [{packageDeps}], [{overrides}]),\n"
+            $"{i}        (\"{package.name}\", \"{package.version}\", \"{package.sha512sri}\", [{sources}], None, [{packageDeps}], [{overrides}]),\n"
```
zhongyi-zhang, line 53: How about exposing a flag to set the specific netrc file path? It could be more friendly than generating `None` to save an extra step updating the generated packages.

zhongyi-zhang:
@purkhusid thank you for improving this PR a lot! Appreciate that you also updated the paket2bazel tool. I tried it with my private registry. Yes, it works. And added some comments.

tools/paket2bazel/Gen.fs
purkhusid: Added a `--netrc-file-label` to the paket2bazel tool. You can now run it with `--netrc-file-label //some:label` and the netrc file will be added to the generated nuget_repo.

purkhusid:
Could you try it again with the flag I added and see how you like it? If you feel like this is a decent solution to this problem I'll go ahead and merge.

zhongyi-zhang:
@purkhusid it all looks good to me. Thank you! Please go ahead merging it.

dotnet/private/rules/nuget/nuget_archive.bzl
```diff
@@ -383,7 +383,7 @@ nuget_archive = repository_rule(
     _nuget_archive_impl,
     attrs = {
         "sources": attr.string_list(),
-        "netrc": attr.string(),
+        "netrc": attr.label(),
```
zhongyi-zhang, line 386: Oh, wait, this change seems to be breaking absolute netrc path usage. I got err like this: `//external:nuget.system.text.json.v7.0.3: invalid label '/tmp/.netrc' in attribute 'netrc' in 'nuget_archive' rule: invalid target name '/tmp/.netrc': target names may not start with '/'`.

zhongyi-zhang: And relative path does not seem to be working, because the default base dir is the cache dir for external pkg:
`ERROR: <my-bazel-cache-dir-for-external-pkg>/microsoft.netcore.app.runtime.linux-x64/7.0.8/BUILD.bazel:5:15: @mdmclient_deps//microsoft.netcore.app.runtime.linux-x64/7.0.8:7.0.8 depends on @nuget.microsoft.netcore.app.runtime.linux-x64.v7.0.8//:data in repository @nuget.microsoft.netcore.app.runtime.linux-x64.v7.0.8 which failed to fetch. no such package '@nuget.microsoft.netcore.app.runtime.linux-x64.v7.0.8//': Not a regular file: <my-bazel-cache-dir-for-external-pkg>/.netrc`

zhongyi-zhang: It works with relative path by putting the netrc file under <WORKSPACE>/external/.

purkhusid: Yes, this is a label to a file within the workspace. So e.g. if you have a `<WORKSPACE>/.netrc` you should put `//:netrc` into this attribute. I wasn't sure if I wanted to allow paths because relative paths won't work since the working directory is always under the output root and thus only absolute paths would work.

purkhusid: Doesn´t the user .netrc fill the need for a global netrc? We default to it if there is not netrc declared in the attribute.

zhongyi-zhang: The `read_user_netrc` method reads `${HOME}/.netrc`. In certain pipeline build scenario, the build tasks run in build agents. Bringing in user-relevant var like `${HOME}` is uncommon.
But I agree that using label to a file is more bazel nature comparing to a string. Probably I should just override `${HOME}` in my scenario.

PR #383, opened 2023-09-19
Make Runfiles library compatible with .NET Standard 2.0

jimevans:
This commit attemps to make the Runfiles library compatible with .NET Standard 2.0 instead of .NET Standard 2.1. The requirement for 2.1 is caused by the use of Path.IsPathFullyQualified(), which only exists for .NET Core 2.0 and later. The implementation of Path.IsPathFullyQualified() has been paraphrased from the code in the .NET runtime source repo.

There may be other ways to accomplish this, including changing which references are included, but this seems the least intrusive mechanism.


purkhusid (approved):
Nice! Thanks for you contribution.

PR #412, opened 2024-02-01
Fix 'repo_mapping_manifest' property check for compatibility

karataliu:
Making it compatible with bazel6.

`repo_mapping_manifest` is added to `FilesToRunProvider` in bazel 7.0 via https://github.com/bazelbuild/bazel/pull/20022

Currently the code will access `repo_mapping_manifest` directly, which makes it only works for bazel >=7.0

This fix helps make the code compatible with bazel 6.0 by adding a attr check. When running on bazel 6.0：
- Before fix: `Error: 'FilesToRunProvider' value has no field or method 'repo_mapping_manifest'`
- After fix: (no error)


purkhusid:
Could you add a comment that explains why this is needed?

karataliu:
@purkhusid  Sure, updated comment in code and PR description. Would you please take another look? Thanks.

purkhusid (approved):

Issue #320, opened 2022-12-06
Download correct app-host for the target RID in order to support cross-compilation


njlr:
Full repro here: https://github.com/njlr/rules-dotnet-publish-docker

```starlark
load("@io_bazel_rules_docker//container:container.bzl", "container_image")
load("@rules_pkg//:pkg.bzl", "pkg_tar")
load("@rules_dotnet//dotnet:defs.bzl", "fsharp_binary", "publish_binary")

fsharp_binary(
  name = "app",
  srcs = [
    "Program.fs",
  ],
  target_frameworks = [
    "net6.0",
  ],
  deps = [
    "@paket.main//microsoft.netcore.app.ref",
    "@paket.main//fsharp.core",
  ],
  visibility = [
    "//visibility:public",
  ],
)

publish_binary(
  name = "app_linux_x64",
  binary = ":app",
  target_framework = "net6.0",
  self_contained = True,
  runtime_identifier = "linux-x64",
  runtime_packs = [
    "@paket.main//microsoft.netcore.app.runtime.linux-x64",
  ],
)

pkg_tar(
  name = "app_linux_x64_archive",
  srcs = [
    ":app_linux_x64",
  ],
  strip_prefix = "app_linux_x64/publish/linux-x64",
  include_runfiles = True,
)

container_image(
  name = "image_debian",
  base = "@dotnet_runtime_deps_6_0_10//image",
  directory = "/home",
  workdir = "/home",
  entrypoint = "/home/app",
  tars = [
    ":app_linux_x64_archive",
  ],
)

publish_binary(
  name = "app_alpine_x64",
  binary = ":app",
  target_framework = "net6.0",
  self_contained = True,
  runtime_identifier = "alpine-x64",
  runtime_packs = [
    "@paket.main//microsoft.netcore.app.runtime.linux-musl-x64",
  ],
)

pkg_tar(
  name = "app_alpine_x64_archive",
  srcs = [
    ":app_alpine_x64",
  ],
  strip_prefix = "app_alpine_x64/publish/alpine-x64",
  include_runfiles = True,
)

container_image(
  name = "image_alpine",
  base = "@dotnet_runtime_deps_alpine_3_16//image",
  directory = "/home",
  workdir = "/home",
  entrypoint = "/home/app",
  tars = [
    ":app_alpine_x64_archive",
  ],
)
```

The Debian based image works fine:

```bash
$ bazel run //:image_debian

$ docker run -it --rm bazel:image_debian                  
[...]

PR #426, opened 2024-04-02
.Net SDK packs

purkhusid:
## What?
Automatic resolution of targeting/runtime/apphost packs. 

## Why?
So that developer experience for the end users is better. They no longer have to think about these packs, which is just like it is in MSBuild.


njlr:
Hi @purkhusid 

Is this PR related to this issue? https://github.com/bazelbuild/rules_dotnet/issues/320

If so, happy to do some testing.


Thanks! :+1:

PR #453, opened 2024-10-03
Fix default transition reset

purkhusid:
## What?
Puts all configuration back to their default values instead of trying to be clever


tomdegoede:
This looks good considering that all build_setting_defaults in [register_tfms.bzl](https://github.com/bazelbuild/rules_dotnet/blame/master/dotnet/private/macros/register_tfms.bzl) are set to False. In our local version of the rules this still is:
```
build_setting_default = sets.contains(TRANSITIVE_FRAMEWORK_COMPATIBILITY[DEFAULT_TFM], framework),
```

tomdegoede (approved):

purkhusid:
Failures look like network errors in CI. Merging

PR #455, opened 2024-10-04
Expose source-files and project structure

njlr:
In order to build Bazel rules for [Fable](https://github.com/fable-compiler/Fable), we need access to the F# source-files and target dependency graph. This information can be used to generate `fsproj` files that the Fable CLI tool can understand. 

This PR exposes these things, but in a generic way, so that it can be used for multiple purposes.

I was able to build a proof-of-concept set of rules for Fable on top of this. However, before going further I would like feedback on these changes, since I hope to eventually have them merged into `rules_dotnet`. 

Thanks!

Related: https://github.com/bazel-contrib/rules_dotnet/issues/315

### DIFF
--- a/dotnet/private/providers.bzl
+++ b/dotnet/private/providers.bzl
@@ -89,3 +89,24 @@ DotnetApphostPackInfo = provider(
         "apphost": "File: The apphost file",
     },
 )
+
+DotnetCompileInfo = provider(
+    doc = "Information about how a .Net target is to be compiled",
+    fields = {
+        "label": "Label: The label of the target",
+        "srcs": "list[File]: sources to be compiled",
+        "deps": "list[DotnetCompileDepVariantInfo]: The direct dependencies of the target",
+        "transitive_deps": "depset[DotnetCompileDepVariantInfo]: The transitive dependencies of the target",
+    },
+)
+
+DotnetCompileDepVariantInfo = provider(
+    doc = "A wrapper provider for a compilation dependency. The dependency can be a project " +
+          "dependency, in which case the `dotnet_compile_info` will be populated" +
+          "or a NuGet dependency, in which case `dotnet_assembly_compile_info` will be populated.",
+    fields = {
+        "label": "Label: The label of the dependency",
+        "dotnet_compile_info": "DotnetCompileInfo: The DotnetCompileInfo of a dependency",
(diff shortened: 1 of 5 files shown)


purkhusid:
I'm not sure about the exact implementation you have for your Fable rules but I wonder why you need to couple the Fable rules with the rules_dotnet ones? AFAIK the Fable compiler only works with the source files so there is not need to compile a dll and then pass the results of that compilation to Fable.

If you can share the fable rules your are working on then I would maybe be able to see the usecase better?

njlr:
> I'm not sure about the exact implementation you have for your Fable rules but I wonder why you need to couple the Fable rules with the rules_dotnet ones? AFAIK the Fable compiler only works with the source files so there is not need to compile a dll and then pass the results of that compilation to Fable.
> 
> If you can share the fable rules your are working on then I would maybe be able to see the usecase better?

The idea is to be able to use `rules_dotnet` targets as inputs into `rules_fable`:

```starlark
load("@aspect_rules_js//js:defs.bzl", "js_binary")
load("@npm//:defs.bzl", "npm_link_all_packages")
load("@npm//:webpack-cli/package_json.bzl", "bin")
load("@rules_dotnet//dotnet:defs.bzl", "fsharp_binary")
load("@rules_fable//fable:defs.bzl", "fable_bundle")

npm_link_all_packages(
  name = "node_modules",
)

fsharp_binary(
  name = "hello_world",
  srcs = [
    "Hello.fs",
  ],
  target_frameworks = [ "net6.0" ],
  deps = [
    "@paket.main//fsharp.core",
  ],
)

fable_bundle(
  name = "fable",
  target = ":hello_world",
)

bin.webpack_cli(
  name = "bundle",
  srcs = [
    ":fable",
  ],
  outs = [
    "dist/main.js",
  ],
  args = [
    "--mode",
    "production",
    "--entry",
    "./fable/Hello.fs.js",
  ],
  log_level = "debug",
)

js_binary(
  name = "app",
  entry_point = ":bundle",
)
```

This means that a user does not need to define the project structure twice.

purkhusid:
I'm not sure I would want to add this as a public interface yet since I don't consider the providers public interfaces yet. This can be solved on the Fable rules side or on the end user side very easily by using a macro.

Sorry if that is a bummer but I'm trying to make the public API surface as small as possible to ease maintenance.

PR #460, opened 2024-11-03
Teach gitattributes/linguist to highlight bzlmod files

petemounce:
(Upstream PR to make this out-of-the-box: https://github.com/github-linguist/linguist/pull/7121)


sin-ack:
Seems like this PR is no longer needed now that Linguist supports this.

purkhusid:
I'm not familiar with this, if this is not needed anymore then I'm going to close. Please repopen if you think this is still needed and I can take a closer look.

PR #470, opened 2025-03-06
Targeting fixes

sin-ack:
These are a couple fixes for when working with non-latest SDKs such as net8.0. We faced these issues when upgrading from `rules_dotnet` 0.15.1 to 0.17.5.


purkhusid:
Do you have an example of what you are exactly fixing with this?

sin-ack:
For the first commit, an example has been given in the commit message: Our projects are targeting net8.0 at the moment, but many packages have upgraded to only support net9.0, and those packages need to fall back to netstandard2.0. Otherwise we'd have to manually pin everything in `paket.dependencies` because many packages in the `Microsoft` namespace have dependency versions >= 8.0 which automatically pull in 9.0 versions. This works the same way in MSBuild.

For the second commit, we have `target_frameworks = ["net8.0"]` for our `csharp_library` targets, but 9.0 versions of targeting pack packages would be pulled in because the targeting pack version 8.0 was being clobbered by version 9.0 in `nuget_repo`.

If you'd like I can try to prepare an example project (and hopefully not disappear this time :sweat_smile:).

purkhusid:
If we could add a test that guards us from a regression then that would be awesome.

sin-ack:
I'll try to do that later today.

sin-ack:
Sorry for the delay, here are the tests.

purkhusid:
Nice! Thanks a lot for your contribution!

purkhusid (approved):

PR #473, opened 2025-03-14
Pass dependencies of analyzers as analyzer DLLs

### LINKED ISSUE #471: Source generators that depend on other libraries/NuGet packages should have their dependencies passed as analyzers
# Problem Definition

I am currently in the process of converting a source generator I wrote in MSBuild to Bazel. The source generator depends on some first-party and some third-party packages.

This is very messy in MSBuild too, see https://github.com/dotnet/roslyn/discussions/47517.

Ideally, this would just work:

```starlark
# Core/BUILD.bazel
csharp_library(
    name = "Core",
    srcs = ["src/**/*.cs"],
    deps = [
        "@paket.main//some.package",
    ],
    target_frameworks = ["net8.0", "netstandard2.0"],
    langversion = "latest",
)

# SourceGenerator/BUILD.bazel
csharp_library(
    name = "SourceGenerator",
    is_analyzer = True,
    srcs = ["src/**/*.cs"],
    deps = [
        # First-party
        "//Core",
        # Third-party
        "@paket.main//microsoft.bcl.asyncinterfaces",
        "@paket.main//microsoft.codeanalysis.csharp",
    ],
    target_frameworks = ["netstandard2.0"],
    langversion = "latest",
)
```

However, I get the following error when I try this:

```
ERROR: /home/agni/[redacted]/Libraries/Foo.SourceGenerator.Tests/BUILD.bazel:9:18: Compiling Foo.SourceGenerator.Tests failed: (Exit 1): compiler_wrapper.sh failed: error executing CSharpCompile command (from target //Libraries/Foo.SourceGenerator.Tests:Foo.SourceGenerator.Tests) external/rules_dotnet~/dotnet/private/compiler_wrapper.sh external/rules_dotnet~~dotnet~dotnet_x86_64-unknown-linux-gnu/dotnet ... (remaining 2 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
warning CS8785: Generator 'FooSourceGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'FileNotFoundException' with message 'Could not load file or assembly 'Microsoft.Bcl.AsyncInterfaces, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.
'.
System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Bcl.AsyncInterfaces, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.

File name: 'Microsoft.Bcl.AsyncInterfaces, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
   at Foo.SourceGenerator.FooSourceGenerator.Execute(SourceProductionContext context, ValueTuple`2 providers)
   at Microsoft.CodeAnalysis.UserFunctionExtensions.<>c__DisplayClass3_0`2.<WrapUserAction>b__0(TInput1 input1, TInput2 input2, CancellationToken token)
-----
```

# Solution

I see two possible solutions:
- Maybe some sort of `analyzer_deps` argument where dependencies to be passed as analyzer assemblies can be explicitly specified.
- If a first-party analyzer library is passed into `deps`, all of its dependencies should also be passed as analyzer assembles. I don't believe this violates `strict_deps` because we already have a well-defined dependency graph.

# Workarounds

For first-party packages, I can create a duplicate target that is `is_analyzer = True`, but for third-party packages no good solution exists (perhaps a custom Bazel rule that extracts the NuGetInfo from third-party packages and moves the libraries to `analyzers`, but we would still need to pass them manually).

sin-ack:
If an analyzer library has dependencies, then those should be passed as analyzers to the compiler. Note that this has to happen during assembly creation and not when another library uses this analyzer, because we need to use the actual runtime libraries present in `DotnetAssemblyRuntimeInfo` as analyzers (not the refs, which is all that's present in `DotnetAssemblyCompileInfo`).

In this change, we only do one level of discovery. This means if a transitive dependency of another library is required in an analyzer library, those must also be passed into deps. This is similar to strict_deps, but is stricter because unlike a regular dependency which only needs its reference DLL to be present at compile time (and reference DLLs don't require their own dependencies), we need the full transitive dependency chain to be present at compile time for an analyzer.

Fixes #471.

TODO:
- [x] Test for ensuring that analyzer deps are passed in as analyzers

### DIFF
--- a/dotnet/private/rules/csharp/actions/csharp_assembly.bzl
+++ b/dotnet/private/rules/csharp/actions/csharp_assembly.bzl
@@ -51,6 +51,24 @@ def _write_internals_visible_to_csharp(actions, label_name, dll_name, others):
 
     return output
 
+def _collect_analyzer_dependencies(deps):
+    """Collect the runtime libraries of this analyzer. These will be passed to the compiler.
+
+    Args:
+        deps: The list of dependencies of the analyzer.
+
+    Returns:
+        The list of analyzer dependencies.
+    """
+
+    dlls = []
+    for dep in deps:
+        runtime_info = dep[DotnetAssemblyRuntimeInfo]
+
+        # FIXME: Should this respect `not strict_deps`?
+        dlls.extend(runtime_info.libs)
+    return dlls
+
 # buildifier: disable=unnamed-macro
 def AssemblyAction(
         actions,
@@ -157,6 +175,12 @@ def AssemblyAction(
         exports,
         strict_deps,
     )
+
+    if (is_analyzer or is_language_specific_analyzer) and target_framework != "netstandard2.0":
+        fail("Analyzers must have `target_frameworks = [\"netstandard2.0\"]`.")
+
+    # TODO: Ensure that all the analyzer DLLs also target netstandard2.0.
+    analyzer_dlls = _collect_analyzer_dependencies(deps) if (is_analyzer or is_language_specific_analyzer) else []
 
     defines = framework_preprocessor_symbols(target_framework) + defines
 
@@ -303,8 +327,8 @@ def AssemblyAction(
         project_sdk = project_sdk,
         refs = [out_ref] if not is_analyzer else [],
         irefs = [out_iref] if out_iref else [out_ref],
-        analyzers = [] if (not is_analyzer) or is_language_specific_analyzer else [out_dll],
-        analyzers_csharp = [out_dll] if is_language_specific_analyzer else [],
+        analyzers = [] if (not is_analyzer) or is_language_specific_analyzer else ([out_dll] + analyzer_dlls),
+        analyzers_csharp = ([out_dll] + analyzer_dlls) if is_language_specific_analyzer else [],
         analyzers_fsharp = [],
         analyzers_vb = [],
         internals_visible_to = internals_visible_to or [],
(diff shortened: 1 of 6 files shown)


dotnet/private/rules/csharp/actions/csharp_assembly.bzl
```diff
@@ -51,6 +51,24 @@ def _write_internals_visible_to_csharp(actions, label_name, dll_name, others):
...
+
+    Returns:
+        The list of analyzer dependencies.
+    """
+
+    dlls = []
+    for dep in deps:
+        runtime_info = dep[DotnetAssemblyRuntimeInfo]
+
+        # FIXME: Should this respect `not strict_deps`?
```
purkhusid, line 68: What does MSBuild do here by default? Does it add the transitive deps as well?

sin-ack: No, it requires the transitive dependencies too, using `TargetPathWithTargetPlatformMoniker`.

purkhusid: Ok, then let's keep it the same here. It can be confusing though so maybe we end up creating a separate rule/macro for analyzer creation since the rules are not really the same as for normal dlls/binaires.

sin-ack:
Is anything else needed here?

PR #475, opened 2025-03-14
Map embedded resource names like MSBuild

### LINKED ISSUE #474: Pass embedded resource names the same way as MSBuild
# Problem Definition

When an `<EmbeddedResource>` item is added to a `.csproj`, the filenames are converted into the format `{project name}.{dotted path}` for the manifest resource in the assembly, where "dotted path" is the relative path of the embedded resource with slashes turned into periods. For example, if `a/b/c.txt` is embedded into the dll for project `Foo.Bar`, the generated name is `Foo.Bar.a.b.c.txt`. This is accomplished by MSBuild passing the resource flag as `/resource:path,name` to csc, where `path` is the actual path to the file and `name` is this generated name.

`rules_dotnet` however only passes the resource as `/resource:path` which turns the name into the last component. So in the example above, the manifest resource is called `c.txt`. This breaks one of my use-cases because I embed a directory by globbing and use the name of the manifest resource to create a virtual file system in memory.

# Solution

I believe MSBuild's approach can be replicated by doing the following:
1. Get the full path of the resource.
2. Strip the repository root of the target from the start.
3. Strip the package path of the target from the start.
4. Replace slashes with periods.
5. Prepend the `out` attribute and a period.

# Open Questions

- [ ] How does this handle paths outside the project folder, or in Bazel's case, potentially cross-repo resources? I'm not sure what MSBuild does here either. We could just fallback to the current way of doing things in this case.

sin-ack:
MSBuild maps the file names as "{project name}.{dotted path}" where the project name is the same as the final generated assembly name and the dotted path is the project-relative path to the embedded resource with slashes changed to periods. Without this, the resource only gets embedded with its basename in the manifest which breaks expectations compared to MSBuild.

Fixes #474.

Drafted since the I haven't received a response on the issue yet. If the solution is acceptable, I will undraft the PR.


sin-ack:
Updated to match what MSBuild does. Undrafting PR as the issue seems to be approved.

purkhusid (approved):
Nice! Thanks for the contribution!

Issue #477, opened 2025-03-17
Building .NET Framework libraries fails: error CS0006: Metadata file 'System.EnterpriseServices.Wrapper.dll' could not be found


sin-ack:
This error seems to happen because of the fix done to solve #405:

https://github.com/bazel-contrib/rules_dotnet/blob/54cf66c0a8d68dd8c5081073da5ac9ce4d762ddd/dotnet/private/rules/nuget/nuget_archive.bzl#L155-L157

However, this causes the following problem:

```
error CS0006: Metadata file '/tmp/bazel-working-directory/_main/external/_main~targeting_packs_extension~nuget.microsoft.netframework.referenceassemblies.net48.v1.0.3/build/.NETFramework/v4.8/System.EnterpriseServices.Wrapper.dll' could not be found
```

However, naively removing this check fails with:

```
error CS1509: The referenced file '/tmp/bazel-working-directory/_main/external/_main~targeting_packs_extension~nuget.microsoft.netframework.referenceassemblies.net48.v1.0.3/build/.NETFramework/v4.8/System.EnterpriseServices.Wrapper.dll' is not an assembly
```

This is because `System.EnterpriseServices.Wrapper.dll` is not an assembly, but a "module", which is a collection of IL code without an assembly manifest (why this is even a thing is an exercise for the reader). It appears that we need another kind of entry beyond "lib" and "ref" called "module", and this library should be added with `-addmodule:`.

lapwingcloud:
Hi @purkhusid I created PR for this issue https://github.com/bazel-contrib/rules_dotnet/pull/506 , could you review and release a new version if reviewed ok ? Thanks .

PR #478, opened 2025-03-17
Implement NET*_OR_GREATER preprocessor defines

sin-ack:
We recursively go through FRAMEWORK_COMPATIBILITY to obtain all of the TFMs we are compatible with and expose them as defines.

.NET Core and .NET Standard flags have been tested. .NET Framework can't be tested because of #477.

### DIFF
--- a/dotnet/private/common.bzl
+++ b/dotnet/private/common.bzl
@@ -507,16 +507,24 @@ def framework_preprocessor_symbols(tfm):
     Returns:
         A list of preprocessor symbols.
     """
-    # TODO: All built in preprocessor directives: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives
 
-    specific = tfm.upper().replace(".", "_")
+    defines = [tfm.upper().replace(".", "_")] + [
+        # net8.0 -> NET8_0_OR_GREATER
+        # net461 -> NET461_OR_GREATER
+        framework.upper().replace(".", "_") + "_OR_GREATER"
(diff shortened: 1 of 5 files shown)


sin-ack:
I'm not quite sure why Windows tests specifically are failing, given that the change shouldn't touch anything related to the tests in question.

dotnet/private/common.bzl
```diff
@@ -507,16 +507,36 @@ def framework_preprocessor_symbols(tfm):
     Returns:
         A list of preprocessor symbols.
     """
-    # TODO: All built in preprocessor directives: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives
 
-    specific = tfm.upper().replace(".", "_")
+    # BFS to recursively find all frameworks that are compatible with the target framework
+    compatible_frameworks = []
+    unvisited_frameworks = [tfm]
+    for _ in range(2147483647):
```
purkhusid, line 514: Can't we find a better way to do this? Not sure what the perf impact during analysis will be if you have a very large repo.

sin-ack: Hmm, well, it's either this or DFS, which will be equally unbounded. We could however pre-compute this.

purkhusid: I think we can reuse this function here: https://github.com/bazel-contrib/rules_dotnet/blob/51e5e038211106296d6eae99bc755faf0e03ccd3/dotnet/private/common.bzl#L18

PR #487, opened 2025-05-12
include assembly version in deps.json

mohaldu:
for context, I am using rules_dotnet with toolchain 9.0.200 to build a binary with both net8.0 and net9.0 and web frameworks. Without including assembly versions in, the net8.0 dll is unable to locate its dependencies


purkhusid (approved):
Nice! Thanks for the contribution!

mohaldu:
Thanks! Can we get this merged and released?

PR #488, opened 2025-05-23
publish_binary: Avoid overwriting files

sin-ack:
I hit this issue when published a binary that included a newer version of an assembly present in the targeting pack (System.IO.Pipelines) as a dependency. The user-provided dependency would be copied, and then the targeting-pack provided version would overwrite it silently. This commit addresses this issue in two ways:

1. Avoid adding the file to the script if it was previously copied.
2. As a failsafe, refuse to overwrite files if they already exist.


sin-ack:
Hmm, seems the build machines lack ASP.NET Core which I required as SDK on the tests since I couldn't find a good assembly to override in just .NET Core. Should that be added, or should I find something in .NET Core to override instead?

sin-ack:
> Runtime pack has lower version than user provided dependnecy

Not sure, will have to test this.

> Runtime pack has higher version than user provided dependency

I know this case since my .NET build works: the user-provided dependency overrides the targeting pack one.

> I'm going on vacation until next week but I can look into this with you after that.

Understood, I'll test the remaining case and let you know. Enjoy your vacation!

purkhusid:
I have a more principled fix in #491 
Thanks for bringing this up and the contribution!

Issue #486, opened 2025-04-29
MODULE.bazel.lock changes across platforms?


njlr:
For local development, I set `lockfile_mode=update` but in CI I have `--lockfile_mode=error`. The idea is that when writing code the `MODULE.bazel.lock` will automatically update and the CI ensures that any changes are maintained in source-control. 

I have noticed that building `rules_dotnet` targets on macOS updates the `MODULE.bazel.lock` file, but not in a way that passes CI, which is Linux. The lock-file updates despite no changes to the Paket files. I can also trigger this by running `bazel mod deps --lockfile_mode update`. 

Is is possible that host platform information is leaking into the lock-file?

Here is an example diff:

```diff
diff --git a/MODULE.bazel.lock b/MODULE.bazel.lock
index 1b08e5034..bbd3f874b 100644
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -174,7 +174,7 @@
   "moduleExtensions": {
     "//deps:paket.main_extension.bzl%main_extension": {
       "general": {
-        "bzlTransitiveDigest": "FxzcrUkOhBab85A9Gm53ngG9XdYLnq4PJoLHvIrgrAY=",
+        "bzlTransitiveDigest": "zBuDjwLL3qV2OXiB3ki6bh86mA54N1kawLnRRx9k6rU=",
         "usagesDigest": "Q4ZbTg4pkfLKcNWiDZnBuToQgG9of8qbypdSAnFioxI=",
         "recordedFileInputs": {},
         "recordedDirentsInputs": {},
@@ -7147,7 +7147,7 @@
     },
     "@@rules_dotnet~//dotnet:paket.paket2bazel_dependencies_extension.bzl%paket2bazel_dependencies_extension": {
       "general": {
-        "bzlTransitiveDigest": "2th3m25C68zC1t81Lo2fo21zdjvKhsTPfse74HjmG5Q=",
+        "bzlTransitiveDigest": "0B+ehe+rqDoMKVI8J6VDJ8+2SC5aJweKx79neKqOVZQ=",
         "usagesDigest": "pat0RBWrcQ4VcWkik5nj60a7ZWuRE9urB6Etrc0pq7g=",
         "recordedFileInputs": {},
         "recordedDirentsInputs": {},
@@ -7778,7 +7778,7 @@
     },
[...]

PR #489, opened 2025-06-02
Mark extensions as reproducible

purkhusid:
## What?
Mark all NuGet extensions are reproducible

## Why?
Limit lockfile churn since all NuGet packages are downloaded using a content hash.


purkhusid:
Fixes #486

Issue #501, opened 2025-07-31
csharp_test on remote workers


tanner0101:
We have remote execution working to build our csharp_binary targets, but we are running into issues running csharp_test.

We get:

```sh
You must install or update .NET to run this application.

App: /buildbuddy/remotebuilds/17f963d3-a61d-4302-a34b-0b04e4344438/bazel-out/...
Architecture: x64
Framework: 'Microsoft.AspNetCore.App', version '9.0.0' (x64)
...

No frameworks were found.
...
```

We tried installing .NET 9 sdk on the worker image per the docs:

> The remote runners do need to have the required .Net system dependencies installed though.

And we can see `dotnet --list-sdks` and runtimes there successfully. 

However, it looks like `DOTNET_ROOT` gets overridden in the test runner: https://github.com/bazel-contrib/rules_dotnet/blob/b169f3e2407deb9dd64678eb0c539e0f94038e7b/dotnet/private/launcher.sh.tpl#L33. It also seems that `PATH` is getting overridden from what we set in our remote runner docker file. 

Is running `csharp_test` remotely supported yet / are we doing something wrong?

Thanks!

purkhusid:
Do you have a copy of your build files that you could provide? Even better would be a minimal reproduction. We already run the test suite on RBE so it is expected to work without installing the .Net SDK on the runners.

tanner0101:
#502 seems to fix the issue for us.

purkhusid:
Fix has been merged on master

PR #502, opened 2025-07-31
Fix remote execution for web sdk (Issue #501)

liembudzien:
Copy over the AspNetCore dependencies, which is needed for remote execution for runnables that require the web sdk instead of default.


purkhusid:
Added your fix and tests in #503 
Thanks for you contribution!

PR #506, opened 2025-08-28
Exclude compiler references to System.EnterpriseServices.Wrapper.dll

lapwingcloud:
This is related to https://github.com/bazel-contrib/rules_dotnet/issues/405 

As mentioned in https://github.com/bazel-contrib/rules_dotnet/issues/477 , I faced the same error like `build/.NETFramework/v4.8/System.EnterpriseServices.Wrapper.dll' could not be found` 

This change moves the exclusion logic before the compliation so that  `System.EnterpriseServices.Thunk.dll` and `System.EnterpriseServices.Wrapper.dll` are not passed as references to compliers but they are still in the inputs. So the compile will work.

## Testing

I have added `e2e/net8.0/basic_csharp_library_net48` folder, which is essentially a copy of `e2e/net8.0/basic_csharp` except the lib's target_frameworks is net48 instead of net8.0, and the ``bazel run hello` works

same for fsharp


purkhusid (changes requested):
Thanks for the contribution! This looks good to me but I would prefer that the tests be in dotnet/private/tests/basic_<csharp/fsharp>_library_net48 since the e2e folders are mainly used for the Bazel registry checks.

purkhusid (approved):

PR #510, opened 2025-09-11
feat: Make the providers part of the public interface.

### LINKED ISSUE #509: Make Providers part of the public interface
I'd like to make the providers part of the public interface, so that they can be used to extend these rules with local extensions.  For example, I use the DotNetAssemblyRuntimeInfo for a `paket pack` implementation, that read the libs and pdbs attributes of that provider.  But to do so I need to disable the visibility check.

Would you prefer that I re-export the providres in dotnet/defs.bzl or move dotnet/private/providers.bzl up the tree?

jaqx0r:
Simply export all the definitions rather than reexporting them through `defs.bzl`
or another starlark file.

Fixes #509

### DIFF
--- a/dotnet/BUILD.bazel
+++ b/dotnet/BUILD.bazel
@@ -109,19 +109,6 @@ bzl_library(
 exports_files(["defs.bzl"])
 
 bzl_library(
-    name = "internal_dev_deps",
-    srcs = ["internal_dev_deps.bzl"],
-    visibility = ["//visibility:public"],
-    deps = ["@bazel_ci_rules//:rbe_repo"],
-)
-
-bzl_library(
-    name = "paket.rules_dotnet_nuget_resource_assemblies_tests_extension",
-    srcs = ["paket.rules_dotnet_nuget_resource_assemblies_tests_extension.bzl"],
-    visibility = ["//visibility:public"],
-)
-
-bzl_library(
     name = "providers",
     srcs = ["providers.bzl"],
     visibility = ["//visibility:public"],
--- a/dotnet/private/tests/appsettings/BUILD.bazel
+++ b/dotnet/private/tests/appsettings/BUILD.bazel
@@ -1,15 +1,3 @@
-load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load(":tests.bzl", "appsettings_test_suite")
 
 appsettings_test_suite(name = "appsettings_test_suite")
-
-bzl_library(
-    name = "tests",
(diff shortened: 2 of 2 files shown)


jaqx0r:
Only what I explained in #509, I don't want to keep setting --nocheck_visibility as I can see by not having it set bugs are creeping in.

My case for exposing the providers now is that if you change how they work, that's ok because you have a versioning scheme, and any smart developer in this modern age is using version pinning, unit tests, and dependabot or renovate.

Eventually however you should make the providers for any public rules also public so that the whole interface is visible to your consumers.

purkhusid:
I wonder why you need to use `--nocheck_visibility`? The providers should be loadable in your code using something like:
```
load("@rules_dotnet//dotnet/private:providers.bzl", "DotnetAssemblyCompileInfo", "DotnetAssemblyRuntimeInfo", "NuGetInfo")
```
Are you getting any errors when you load them from this location?

jaqx0r:
I'm on leave at the moment, so I can't verify immediately, but yes I am pretty sure I had visibility errors trying to load those symbols.  Let me get back to you next week with confirmation though.

jaqx0r:
The import line in my bzl file is:

```
load("@rules_dotnet//dotnet/private:providers.bzl", "DotnetAssemblyRuntimeInfo")
```

You're right, there is an error in my BUILD.bazel where I was _also_ depending on the `common` target:

```
bzl_library(
    name = "paket",
    srcs = ["paket.bzl"],
    visibility = ["//visibility:public"],
    deps = [
        "@aspect_bazel_lib//lib:paths",
        "@rules_dotnet//dotnet/private:common",  # oops
        "@rules_dotnet//dotnet/private:providers",
    ],
)
```

But removing that line I still get the same error:

```
ERROR: /[redacted]/dotnet/bzl/BUILD.bazel:9:12: in bzl_library rule //dotnet/bzl:paket: Visibility error:
target '@@rules_dotnet+//dotnet/private:providers' is not visible from
target '//dotnet/bzl:paket'
Recommendation: modify the visibility declaration if you think the dependency is legitimate. For more info see https://bazel.build/concepts/visibility
```

https://github.com/bazel-contrib/rules_dotnet/blob/master/dotnet/private/BUILD.bazel#L104-L108 is private to the package.

Maybe I misinterpreted what you meant with "loading the `bzl_library` target"  ?

purkhusid:
I guess that I'm just confused why you are creating a `bzl_library` target. Do you have a specific use case for that?

PR #511, opened 2025-09-30
Fix copy of appsetting_files and detect namespace of resources

peakschris:
This PR fixes two things:

1) appsettings.json files are not always copied to output directory. This is fixed by adding the appsettings outputs to data.
2) improve error handling related to failures to copy appsettings
3) resources named with fully qualified namespace - as suggested by resgen.exe - are not added to the compile line correctly.

Resgen promotes naming the output .resources file with fully qualified namespace such as Company.Department.Class.resources; when files that contain such a dot-qualified namespace are provided use the filename namespace directly rather than attempting to derive it from out_dll.


purkhusid (commented):
This looks good to me, any chance you could add a test that captures this under `dotnet/private/tests/resources`?

PR #512, opened 2025-10-04
Avoid duplicate analyzers

peakschris:
For targets that require both net8 (default) and deps = ["@paket.main//microsoft.aspnetcore.app.ref"] duplicate analyzers are configured which results in compilation errors.

Avoid duplicate analyzers based on their filenames.

I'm not particularly happy with this fix; but it works for my case; happy to take suggestions for improvement.

Chris


purkhusid:
This should be fixed on master

PR #529, opened 2026-02-02
Update paket2bazel README

J-B-Blankenship:
Add Bazel MODULE instructions. Point new users explicitly to `paket2bazel` for external dependency management.

This took 8+ hours to backwards engineer. To be fair, I do not develop with C#, so the nuances were lost on me for many steps. The changes here are targeting folks new to the ecosystem with explicit instructions vs. having to dig into the repo.

The update should hopefully trigger search assistants such as Google AI, DuckDuckGo AI, and local models to give correct feedback on issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,3 +14,7 @@ primer on Bazel.
 See API documentation at https://registry.bazel.build/docs/rules_dotnet
 
 See examples in the [examples](examples/) folder.
+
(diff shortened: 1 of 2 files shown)


purkhusid:
paket2bazel is not needed anymore. New and improved Paket handling is now on master.

PR #535, opened 2026-03-23
Fix regression vs. 0.19.2 in local native library support

b-spencer:
This PR fixes what I believe to be a regression vs. 0.19.2 with respect to native libraries that do not come from NuGet packages.

In `rules_dotnet` 0.21.5, attempting to use a `csharp_import_library()` for a locally-built `cc_library()` shared library via a `csharp_nunit_test()` fails because the shared library can’t be found.  There appear to be other cases other than NUnit tests that fail, too.

This happens because the `.deps.json` file generated for the unit test tries to specify a NuGet-style `runtimes/...` path for the shared library, but doesn’t place the library in such a location.  For example, with 0.21.5, the `.deps.json` might look like this for a locally-built native library:

```json
                        "native/": {
                                "dependencies": {},
                                "resources": {},
                                "runtime": {},
                                "runtimeTargets": {
                                        "runtimes/tests/local_native_library/libnative.so": {
                                                "assetType": "native",
                                                "rid": "tests"
                                        }
                                }
                        }
```

Note how the RID search algorithm has assumed that the path to the library is one from a NuGet package and it has selected various parts of the path to the actual library within the `bazel-out/` tree as if they were parts of a `runtimes/RID/native/...` style path.

Ignoring the RID generation issue, it seems that what should happen is that the `.deps.json` should be generated as it was for locally-built native libraries in `rules_dotnet` 0.19.2.  This PR fixes this behaviour so that once again the relevant section of `.deps.json` looks like this:

```json
                        "native/": {
                                "dependencies": {},
                                "native": {
                                        "_main/dotnet/private/tests/local_native_library/libnative.so": {
                                                "fileVersion": "0.0.0.0"
                                        }
                                },
                                "resources": {},
                                "runtime": {}
                        }
```

Note how it’s simply a native library with a path relative to the execroot (which itself is part of the search path via other arrangements that `rules_dotnet` sets up).

The fix in this PR is a small change in `dotnet/private/common.bzl`.  There’s a condition inside `generate_depsjson()` that is essentially missing a case for native libraries that don’t come from NuGet packages.  If we emit (almost) the same `.deps.json` content as 0.19.2 did, things work again.

I first encountered this with unit tests, but it might affect other target types.  This PR adds tests that produce a `csharp_binary()` which seems also broken without this fix.  A published self-contained program seems to work fine with this fix, too.  However, it's possible that I've missed some case.  (All the `rules_dotnet` tests pass in the devcontainer.)

BTW, I don’t know if the RID path generation that exists elsewhere in `rules_dotnet` should be fixed or changed, too.  It is still producing the invalid paths, but the change I show here effectively side-steps that by ignoring its output.

The first commit reproduces the issue without the fix, and the second commit fixes it.


purkhusid:
Ok, using C/C++ libraries in rules_dotnet sounds like a great addition but I wonder if we could change the approach taken here to so that the end user can just omit the import library part and instead just add the `cc_shared_library` directly to deps. rules_dotnet could be updated to allow deps that provide the Providers that `cc_shared_library` provides and then those providers could be handled separately from the normal .Net dependencies.

b-spencer:
Interesting.  I didn't realize it wasn't already a supported feature.  We've been using 0.19.2 and doing exactly like the example shows and everything has been working so far (on Windows, macOS, and Linux).

As a result, I've had to apply this patch locally to keep things working after upgrading :)

b-spencer:
Maybe it's possible to allow "clumsy" support like this until there is time to do it better?

purkhusid:
@b-spencer I changed the test structure a bit and it surfaced that the current approach does not work for framework dependent publishing. I think this is due to the naive assumptions that I made here: 
https://github.com/b-spencer/rules_dotnet/blob/66ee1d1013ff246c802ab427351f47a323a7a08e/dotnet/private/common.bzl#L652

and here: 
 https://github.com/b-spencer/rules_dotnet/blob/66ee1d1013ff246c802ab427351f47a323a7a08e/dotnet/private/rules/publish_binary/publish_binary.bzl#L85

I don't have time to figure out a solution currently but I think the approach that you have take in the PR is acceptable until we have some kind of first party support for the rules_cc providers. If you have the time to figure out a solution to the framework dependent issue then I can merge this PR once that has been solved.

b-spencer:
How's this for an approach?  I don't _think_ the assumption is too tenuous, but maybe I've overlooked something.  I considered doing this only when the potential RID value extracted from the path wasn't a match for the `RUNTIME_GRAPH` keys in `rids.bzl` but that somehow felt more hacky and ultimately not necessary.

purkhusid:
I think the approach is not worse than the one before it. A more disciplined approach requires quite a bit of work so I think this will do for now.

PR #540, opened 2026-06-08
Fix filepaths with space

soda0289:
I found an issue where if a file path has a space "foo/api tests/some_test.cs" then the compile would fail. This seems to only happen with csharp compiler. I think the fsharp compiler sends each cli argument as a new line which makes it support spaces in folder and file names.

I can add more tests as well to test each possible case. First time looking into this code base maybe there's a better way to handle it as well.


purkhusid:
Thanks for the fix! I think the e2e tests are not necessary since they are more of a smoke test. If you could remove the e2e tests then I'll merge the fix!

soda0289:
Remove e2e tests and added a few more unit tests for resources and references with whitespace.

purkhusid:
Don't worry about the failed Buildifier. I'll fix that after I merge. 
One more thing. Does F# fail with the same path handling? If not then I would prefer that the path handling just also be applied there to keep the differences between C# and F# as small as possible.

soda0289:
~I applied the same whitespace handling to F#. There is no issue adding it in and it also allows us to move the format_arg functions into common.bzl in private and remove some conditions that were only for csharp. Let me know if I should change anything else.~

soda0289:
After doing more testing I found that the fsharp compiler doesn't strip the double quotes it includes them in the output. Will revert change and add tests for fsharp to assert no quotes are added.

soda0289:
Pushed the change to ensure F# compiler doesn't add quotes. Quotes do work for source file but not anything else.

purkhusid:
I updated the PR to use the built in Bazel args formatting. This should take care of it. I also simplified the testing since I think it´s enough to test that the build works with DLLs containing spaces.

purkhusid:
Thanks for the contribution!

soda0289:
Thanks for cleaning it up I appreciate it.

PR #541, opened 2026-06-09
Add env_inherit attribute for tests

soda0289:
This PR adds the [common test attribute](https://bazel.build/reference/be/common-definitions#common-attributes-tests) `env_inherit` that is found on other test libraries such as `cc_test`, `py_test`, and `sh_test`. 

This attribute allows for testing situations where the users local machine or CI needs to pass in credentials. For example if running tests using Testcontainers that has to pull down an image from a private registry, `env_inheirt` can be used to pass in the docker config location which will be based on their home directory `$HOME/.docker/config.json` and not rely on hard coding the path, which would be different for each user, if just the `env` attribute was used. We could also use `--test_env` to solve this but that would mean every test would get access to this variable where only ones that run a test container need it.

### DIFF
--- a/dotnet/private/rules/common/attrs.bzl
+++ b/dotnet/private/rules/common/attrs.bzl
@@ -138,9 +138,9 @@ COMMON_ATTRS = {
 LIBRARY_COMMON_ATTRS = {
     "exports": attr.label_list(
         doc = """
-        List of targets to add to the dependencies of those that depend on this target. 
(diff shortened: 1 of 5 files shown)


purkhusid (approved):
I like it! Thanks for the contribution!

purkhusid:
I'll fix the CI after the merge. I need to update the Bazel version and buildifier and more.

PR #543, opened 2026-06-16
Include appsetting_files in runfiles

FrankPortman:
appsetting_files end up in DefaultInfo.files but not in runfiles, so they're missing when a target runs from an isolated runfiles tree (remote execution / sandbox). Locally everything's next to the assembly so it works; on RBE the test fails with FileNotFound for filesettings.json.

One-line fix: seed additional_runfiles with appsetting_files.

Assuming this is just an oversight, but happy to close if appsetting files are intentionally kept out of runfiles.


purkhusid:
Thanks for the contribution. I'll fix the CI now.

Issue #467, opened 2025-01-24
Source generator issue in Microsoft.Configuration.Binder package


leryss:
We are still getting source generator this issue that is discussed in this issue https://github.com/bazel-contrib/rules_dotnet/issues/443
We are using version 0.17.5 and according to the issue above the issue should be fixed.

```
Microsoft.Extensions.Configuration.Binder.SourceGeneration/Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator/BindingExtensions.g.cs(38,10): error CS0433: The type 'InterceptsLocationAttribute' exists in both 'Common.Kafka, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' and 'Common.Kafka, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'
Microsoft.Extensions.Configuration.Binder.SourceGeneration/Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator/BindingExtensions.g.cs(38,10): error CS0433: The type 'InterceptsLocationAttribute' exists in both 'Common.Kafka, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' and 'Common.Kafka, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'
```

Since you've already worked on this issue can you provide some information what causes this ? I don't have a reproducible example but atleast i would like to understand what causes this.

purkhusid:
Do you have a reproduction of this?

soda0289:
This source generator `Microsoft.Extensions.Configuration.Binder.SourceGeneration.dll` is included twice from the nuget package `Microsoft.Extensions.Configuration.Binder` and in the sdk web asp.net framework. By default these analyzers are not included but since rules_dotnet includes all analyzers this error can happen. To trigger the error you need to use the configuration binder api (` .Get<T>()` ,  `.Bind()`,  `Configure<T>(config)`) and the nuget version of `Microsoft.Extensions.Configuration.Binder` analyzer is newer than the version the baseline rules_dotnet records in [targeting_pack_overrides](https://github.com/bazel-contrib/rules_dotnet/blob/63451d5772e28758460b57913c151278be79ce96/dotnet/private/common.bzl#L266).

[...]

PR #549, opened 2026-07-14
Add excluded_analyzers attribute

soda0289:
This PR adds an option `excluded_analyzers` that allows an analyzer from a nuget package to not be run. By default `run_analyzers` is true which runs all analyzers. Some nuget packages define a `.targets` msbuild file that can disable/remove an analyzer.

Nuget example:
```xml
    <Target Name="_Microsoft_Extensions_Configuration_Binder_RemoveAnalyzer"
            Condition="'$(EnableConfigurationBindingGenerator)' != 'true'"
            AfterTargets="ResolvePackageDependenciesForBuild;ResolveNuGetPackageAssets">

        <ItemGroup>
            <Analyzer Remove="@(Analyzer->WithMetadataValue('NuGetPackageId', 'Microsoft.Extensions.Configuration.Binder'))" />
        </ItemGroup>
    </Target>
```

This example is from [Microsoft.Extensions.Configuration.Binder nuget package](https://nuget.info/packages/Microsoft.Extensions.Configuration.Binder/10.0.9). It contains `Microsoft.Extensions.Configuration.Binder.targets` file in the buildTransitive folder that removes its analyzer unless a condition `EnableConfigurationBindingGenerator` is set to true.

One caveat is that `excluded_analyzers` will not fail or error out if the analyzer is not included since some analyzers are target framework specific and might be in one tfm but not another and I couldn't figure out a way to handle that nicely. I could add back in a fast fail() if it isn't included but then users would be forced to use select() to optionally add the excluded_analyzers option per target framework.

In the future it might be nice to actually parse these msbuild target files but it would add a lot of complexity and if we want to match what msbuild is doing with nuget packages adding a `excluded_analyzers` attribute seems like a good option.


soda0289:
This solution could also help with issues like #467 since if a source generator is included when it should not be and the source generator is loaded more than once, it will crash and not behave the same way as msbuild or dotnet build would. By excluding the `Microsoft.Extensions.Configuration.Binder.SourceGeneration.dll` source generator/analyzer dotnet falls back to the runtime assembly analyzer. Which is the default behavior of msbuild since both the [Microsoft.Extensions.Configuration.Binder nuget package](https://nuget.info/packages/Microsoft.Extensions.Configuration.Binder/10.0.9) and [sdk framework](https://github.com/dotnet/sdk/blob/5ef6b4b9d2a2de6ad3a6e3b2e9ef64e5ba08ed8e/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets#L501) both disable the analyzer by default.

Issue #448, opened 2024-09-07
Can't use nuget packages with no framework versions specified


0x2Adr1:
Example: `WebSocketSharp`: https://nuget.info/packages/WebSocketSharp/1.0.3-rc11

The generated BUILD file is missing `websocket-sharp.dll`

```
package(default_visibility = ["//visibility:public"])
exports_files(glob(["**"]))
load("@rules_dotnet//dotnet/private/rules/nuget:nuget_archive.bzl", "tfm_filegroup", "rid_filegroup")
filegroup(name = "libs", srcs = [])
filegroup(name = "refs", srcs = [])
filegroup(name = "analyzers", srcs = [])
filegroup(name = "data", srcs = [])
filegroup(name = "native", srcs = [])
filegroup(name = "content_files", srcs = [])
exports_files(["websocketsharp.1.0.3-rc11.nupkg"])
```

As a consequence depending on `@paket.main//websocketsharp` has no effect:

```
fsharp_binary(
    name = "foo",
    srcs = ["foo.fs"],
    deps = ["@paket.main//websocketsharp"],
)
```

```
error FS0039: The namespace or module 'WebSocketSharp' is not defined.
```

0x2Adr1:
I made it work with this diff, far from perfect though:

```
> git diff dotnet/private/rules/nuget/nuget_archive.bzl
diff --git a/dotnet/private/rules/nuget/nuget_archive.bzl b/dotnet/private/rules/nuget/nuget_archive.bzl
index bd25e52..b9ad6fa 100644
--- a/dotnet/private/rules/nuget/nuget_archive.bzl
+++ b/dotnet/private/rules/nuget/nuget_archive.bzl
@@ -108,7 +108,21 @@ def _process_group_with_tfm(groups, group_name, file):
     i = file.find("/")
     tfm_start = i + 1
     tfm_end = file.find("/", i + 1)
-    tfm = _replace_non_standard_tfm(file[tfm_start:tfm_end])
+    tfm_name = file[tfm_start:tfm_end]
+    tfm = _replace_non_standard_tfm(tfm_name)
+
+    group = groups[group_name]
+
+    # SPECIAL CASE FOR PACKAGES THAT HAVE THEIR DLLs DIRECTLY in lib/ LIKE WebSocketSharp
+    if tfm_end == -1:
+        filename = file[tfm_start:]
+        if filename.endswith(".dll"):
+            tfm = "netstandard2.0"
[...]

PR #552, opened 2026-07-31
fix(tfm_filegroup): always create _default target referenced by select()

brainoreally:
## Problem

tfm_filegroup() unconditionally references :<name>_default as the //conditions:default fallback in multiple select() calls, but only created that target when default_to_empty=True. This causes "missing input file" errors when:

- Using //... wildcard builds (e.g., bazel build //...)
- Running in multi-platform configs where no TFM condition matches and the default case is evaluated

## Root Cause

In nuget_archive.bzl, the _default empty filegroup was wrapped in:

```python
if default_to_empty:
    native.filegroup(name = "%s_default" % name, srcs = [], ...)
```

However, default_to_empty=False is passed for libs and refs groups, yet their aliases still reference :<name>_default unconditionally. The guard makes the fallback conditional while the references are not.

## Fix

Remove the guard — always create _default regardless of default_to_empty.

This is safe because _default is always an empty filegroup (srcs=[]), serving only as a no-op fallback. Making it unconditional:

- Fixes the build failure
- Has zero semantic change to consumers (no extra files are selected)
- Aligns with how the fallback was already unconditionally referenced

## Related

- Same issue as #448 (closed without upstream fix merged)
- Reproduced on both main and v0.20.1

## Verification

Verified locally via downstream project (bazel build //...) and will be exercised by repo CI.

**Cc:** @purkhusid for review — primary maintainer of the nuget code.

### DIFF
--- a/dotnet/private/rules/nuget/nuget_archive.bzl
+++ b/dotnet/private/rules/nuget/nuget_archive.bzl
@@ -585,12 +585,11 @@ def tfm_filegroup(name, default_to_empty, tfms):
     cor = []
     tfm_rids = {}
 
-    if default_to_empty:
-        native.filegroup(
-            name = "%s_default" % (name),
-            srcs = [],
-            visibility = ["//visibility:public"],
-        )
+    native.filegroup(
+        name = "%s_default" % (name),
+        srcs = [],
+        visibility = ["//visibility:public"],
+    )
 
     for (tfm, value) in tfms.items():
         native.filegroup(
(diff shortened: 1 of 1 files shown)


purkhusid:
I'm not sure about this change since this just moves the error to runtime. The linked issue #448 was closed due not wanting to support NuGet packages that are not authored correctly.

If you have an exact example of the problem you are solving then I would prefer that you add a regression test to show it.

purkhusid:
@brainoreally Could you check the latest master and see if it solves your issue? I patched some gaps in our NuGet resolution. If that does not work then I would appreciate the name of the NuGet package that is causing you trouble.

PR #570, opened 2026-09-16
Provisional Blazor support for rules_dotnet

### LINKED ISSUE #249: Razor/Blazor Support
I would like to build a small blazor server using bazel. The server itself builds just fine, but ends up broken, because razor, cshtml.cs and various data files are not added to the target as required.

Is building blazor servers supported at all? If yes, may there be an example somewhere on how to build those?

If not presently, may the support for those be added in the near enough future?

oakad:
These changes allow the use of rules_dotnet for building Blazor WASM apps. I have focused on WASM because of my personal needs, however, extention to other Blazor models may be reasonably straightforward. Provisions are made to support localization and asset embedding.

Fixes #249.

Caveat: this PR is just an implementation proposal, albeit working.


purkhusid:
I've added the initial Razor/Blazor support on master. Would appreciate if you would take a look and try it out and see if it is lacking any features you currently use.

